PR #6, opened 2026-05-13
feat: inline tab renaming and inline project renaming from command palette

jcuhnoio:
## Summary

- **New**: Tab renaming — tabs can now be renamed inline directly in the sidebar, triggered from the context menu or the command palette ("Rename current tab")
- **Changed**: Project renaming from the command palette no longer opens a modal dialog; it now uses the same inline sidebar editing as the context menu
- **Fixed**: Auto-renaming is preserved — pressing Enter without changing the name is a no-op, and clearing the field restores the process-derived title
- **UX**: Rename field placeholder shows the auto-generated title so the user always knows what the tab would revert to
- **UX**: Opening the sidebar is automatic when a rename is triggered while it is hidden

## How it works

Both rename flows share the same architecture:

- `AppState` exposes `renamingTabID` and `renamingProjectID`; setting either triggers the corresponding sidebar row to enter inline edit mode
- A `postPaletteAction` hook on `AppState` lets palette actions register a callback that runs after the palette has fully dismissed — this is the general solution to the focus-competition problem between the palette and any sidebar TextField, and replaces the earlier per-feature special-casing in `MainWindow`
- `restoreFocusToActivePane()` was moved from `MainWindow` into `AppState` so the sidebar rows can call it on submit/escape without needing to know about `MainWindow`
- `TerminalTab` now exposes `autoTitle` (process-derived, no custom override) alongside `sidebarTitle` (`customTitle ?? autoTitle`), so the rename field can show the correct placeholder and the serializer can round-trip custom titles independently

## Test plan

- [x] Rename a tab via sidebar context menu → inline field appears, Enter commits, Escape cancels, focus returns to terminal
- [x] Rename a tab via command palette ("Rename current tab") → palette closes, sidebar opens if hidden, inline field appears
- [x] Rename a project via sidebar context menu → same inline behaviour as before
- [x] Rename a project via command palette ("Rename project") → inline field instead of modal dialog
- [x] After any rename (commit or cancel), cursor/focus is back in the terminal without a mouse click
- [x] Pressing Enter without typing anything leaves the auto-title intact
- [x] Clearing a custom name and submitting restores the process-derived auto-title
- [x] Rename field placeholder shows the auto-generated title, not the current custom name


thdxg:
@jcuhnoio thanks for your contribution 🚀 Can you address the formatting, linting, and test failures?

thdxg:
I like this idea generally. Some feedbacks:
- If the sidebar is collapsed, I think it should expand when the user tries to rename a tab/project.
- The tab auto-renaming seems to be broken now. Unless the user renames the tab, it should always update to the running process's name. (e.g., if you open nvim from zsh, then the tab name should change from zsh to nvim)

jcuhnoio:
> Can you address the formatting, linting, and test failures?

I believe they have been resolved. Along with that, a couple tests have been added for the implemented feature.

> If the sidebar is collapsed, I think it should expand when the user tries to rename a tab/project.

Do you think it should collapse once the user is done renaming? The expand feature has been added.

> The tab auto-renaming seems to be broken now. Unless the user renames the tab, it should always update to the running process's name. (e.g., if you open nvim from zsh, then the tab name should change from zsh to nvim)

Resolved.

thdxg:
> Do you think it should collapse once the user is done renaming? The expand feature has been added.

I think current state is good. It lets the user confirm the changed tab name.

> > The tab auto-renaming seems to be broken now. Unless the user renames the tab, it should always update to the running process's name. (e.g., if you open nvim from zsh, then the tab name should change from zsh to nvim)
> 
> Resolved.

It seems like test and format failed. You can run the ci workflows locally with `mise run ci:test` and `mise run ci:format`.
You can also format with `mise run format`.

jcuhnoio:
## ci:format lint failure on GeneratedAssetSymbols.swift (local only)
  
After fixing the redundantThrows violation in the test file, ci:format still fails locally due to GeneratedAssetSymbols.swift, a file Xcode generates into build/DerivedData/ at build time:

```
  build/DerivedData/.../GeneratedAssetSymbols.swift:16:1: error: (emptyBraces)
  build/DerivedData/.../GeneratedAssetSymbols.swift:17:1: error: (blankLinesAtEndOfScope)
  ...
```

This seems to be a local environment issue — ci:format passes on GitHub Actions, presumably because DerivedData isn't present in a clean CI checkout. The .swiftformat config already excludes .build/ and GhosttyKit/, but the project's DerivedData lands under build/ (no leading dot), so SwiftFormat picks it up locally.

thdxg:
lgtm! thanks for your contribution~

Issue #2, opened 2026-04-14
Dock icon size is bigger


wildmildnull:
<img width="265" height="115" alt="Image" src="https://github.com/user-attachments/assets/a55b5686-4373-4187-8e8a-1f1a7dcee3b0" />

### Description

Dock icon size is bigger

### Steps to reproduce

Dock icon size is bigger

### Expected behavior

Normal as usual.

### Macterm version

v1.2.1

### macOS version

macOS 26.2

### Logs or screenshots

Normal

thdxg:
This is fixed as of `v1.2.2` 🫡

Issue #1, opened 2026-04-14
Vertical two windows mixed up the mouse click event


wildmildnull:
### Description

As title said.

### Steps to reproduce

Window A and Window B after split down a window
Mouse click Window A, but Window B gets focus.

### Expected behavior

Click window A, it got focus.

### Macterm version

v1.2.1

### macOS version

macOS 26.2

thdxg:
Thanks for reporting! The issue has been resolved now. It will be fixed in the release `v1.2.2`

PR #19, opened 2026-05-19
feat: desktop notifications

hrbrmstr:
Add macOS desktop notifications for terminal events:

- **OSC 7770** notifications (`GHOSTTY_ACTION_DESKTOP_NOTIFICATION`) — explicit notification escape sequences from shell programs
- **Command finish** notifications (`GHOSTTY_ACTION_COMMAND_FINISHED`) — fired by `notify-on-command-finish` config with exit code and human-readable duration

Notifications are only shown when the app is unfocused or the source pane isn't key. Clicking a notification navigates to the source pane (activates app, switches project/tab, focuses pane). Quick Terminal clicks show the panel instead.

Key changes:
- `Pane` gains `projectID` for click-to-navigate routing
- `NotificationHandler` — `UNUserNotificationCenterDelegate` with authorization and click routing
- `AppState.navigateToPane` — workspace/tab/pane navigation from notification clicks
- `QuickTerminalService.showPanel()` — show-only (never hides) for notification clicks
- Both `onDesktopNotification` and `onCommandFinished` callbacks nil'd in `destroySurface`

Requires ghostty config:
```
notify-on-command-finish        = unfocused
notify-on-command-finish-action = notify
desktop-notifications           = true
```


thdxg (commented):
Thanks for your contribution! A few things to address before merging:

### Blocking

**1. `navigateToPane` bypasses `selectProject` / `ensureWorkspace`** (`Macterm/App/AppState.swift:304-321`)

The plan explicitly called for using `selectProject` "rather than setting `activeProjectID` directly" so workspace setup and recency stay consistent. The current code does `recordProjectVisit(projectID); activeProjectID = projectID` without `ensureWorkspace`. In practice the workspace will exist (a notification was posted from a live pane, so the workspace was created at app start or via `selectProject` earlier), but if a project was deleted while a notification was queued, the click can leave `activeProjectID` pointing at a missing workspace. Either:

- Look up the `Project` from `ProjectStore` and call `selectProject(_:)`, falling back to "just activate the app" if it's gone — matches the plan's "if pane no longer exists, skip and activate"; or
- At minimum guard `workspaces[projectID] != nil` before assigning `activeProjectID`.

**2. `NotificationHandler.swift:25-37` re-reads `userInfo["paneID"]` instead of using the `paneID` local**

```swift
if let focusedIDString = userInfo["paneID"] as? String,
   let focusedID = UUID(uuidString: focusedIDString),
```

`paneID` is already in scope from the guard above. Drop the re-extraction and use `paneID` directly. Not a bug, but the duplication makes it look like there's a separate "focused pane ID" concept that doesn't exist.

### Non-blocking suggestions

**3. `QuickTerminalService.projectID` is a per-process static**

Fine because notifications don't outlive the process, but worth a one-line comment so the next reader doesn't add code that assumes persistence (e.g. comparing across restarts).

**4. Notification identifier collision risk** (`TerminalPane.swift:144-148`, mirrored in `onCommandFinished`)

`Date().timeIntervalSince1970` is a `Double` and two notifications fired in the same runloop tick from the same pane would format to the same identifier. macOS would coalesce the second into the first. `UUID().uuidString` is a stricter unique-id and the same length cost.

**5. Authorization callback discards the granted flag** (`NotificationHandler.swift:9`)

```swift
UNUserNotificationCenter.current().requestAuthorization(options: [.alert]) { _, _ in }
```

If the user denies, every later `UNUserNotificationCenter.add` silently no-ops. That's the documented macOS behavior and probably what we want, but logging the denial once (or threading the `granted` flag onto `NotificationHandler` so a future Settings UI can surface the state) would save someone debugging "why no notifications" later. Low priority.

**6. `formatDuration` rounds 59.9s → "59.9s" but 60s → "1m 0s"**

Cosmetic — the boundary is fine. No change needed; just flagging that the `Int(seconds)` truncation in the `mins/secs` branch will display "0m 59s" for anything ≥ 60 but < 60.5, which is a sub-second edge with no real impact.

**7. Planning docs in-tree** (`docs/brainstorms/...`, `docs/plans/...`)

384 lines of planning markdown in the repo is unusual for this codebase. The PR description already conveys the design. Consider dropping these.


Happy to take another pass once 1 and 2 are addressed.

hrbrmstr:
Addresses review feedback from #1:

**Blocking fixes:**
- `navigateToPane` now guards `workspaces[projectID] != nil` before setting `activeProjectID`; falls back to just activating the app if workspace is missing
- `NotificationHandler` uses `paneID` directly instead of re-extracting from `userInfo`

**Non-blocking improvements:**
- Notification identifiers use `UUID().uuidString` instead of `Date().timeIntervalSince1970` to avoid same-tick collision
- Log notification authorization denial once via `os.Logger`
- Remove planning docs from git tracking, add `docs/brainstorms/` and `docs/plans/` to `.gitignore`

**Skipped:**
- Fix 3 (comment on `QuickTerminalService.projectID`): project convention prohibits comments
- Fix 6 (`formatDuration` boundary): reviewer flagged as cosmetic, no change needed

hrbrmstr:
Addresses remaining review items from #1 and #2 that were previously skipped:

- **Fix 3**: Rename `QuickTerminalService.projectID` → `ephemeralProjectID` to convey the per-process nature self-documentingly (no comment needed per project convention)
- **Fix 6**: Round seconds before truncating in `formatDuration` to avoid displaying "0m 59s" for durations just above 60s

thdxg:
@hrbrmstr I tested this feature locally and your PR looks good! Thanks for your contribution!

PR #28, opened 2026-05-25
feat: paste copied file paths into terminals

m0n99:
## Summary

- Paste Finder-copied files as shell-escaped paths
- Use the same pasteboard handling for Cmd+V and context menu Paste
- Helps tools like opencode accept pasted screenshots/images by receiving the copied file path

## Verification

- mise run format
- mise run lint
- mise run test

### DIFF
--- a/Macterm/Ghostty/GhosttyCallbacks.swift
+++ b/Macterm/Ghostty/GhosttyCallbacks.swift
@@ -58,9 +58,75 @@ final class GhosttyCallbacks: @unchecked Sendable {
     }
 
     func readClipboard(ud: UnsafeMutableRawPointer?, location: ghostty_clipboard_e, state: UnsafeMutableRawPointer?) -> Bool {
-        let text = NSPasteboard.general.string(forType: .string) ?? ""
+        let text = Self.readPasteboardText() ?? ""
         text.withCString { ghostty_surface_complete_clipboard_request(surface(from: ud), $0, state, false) }
         return true
+    }
+
+    // MARK: - Pasteboard text resolution (shared with GhosttyTerminalNSView)
+
+    /// Mirrors Ghostty's ShellEscapeWriter for file paths pasted into the shell.
+    private static func shellEscape(_ s: String) -> String {
+        var result = ""
+        result.reserveCapacity(s.utf8.count)
+        for ch in s {
+            switch ch {
+            case "\\",
+                 "\"",
+                 "'",
+                 "$",
+                 "`",
+                 "*",
+                 "?",
+                 " ",
+                 "|",
+                 "(",
+                 ")":
+                result.append("\\" + String(ch))
+            default:
+                result.append(ch)
+            }
+        }
+        return result
+    }
+
+    /// Returns pasted text from the pasteboard: file paths (Finder drag/copy)
+    /// fall back to plain string. Called by both the context-menu paste path
+    /// and the libghostty Cmd+V clipboard path.
+    static func readPasteboardText() -> String? {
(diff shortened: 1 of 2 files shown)


Macterm/Ghostty/GhosttyCallbacks.swift
```diff
@@ -58,11 +58,77 @@ final class GhosttyCallbacks: @unchecked Sendable {
...
+                 "\"",
+                 "'",
+                 "$",
+                 "`",
+                 "*",
+                 "?",
+                 " ",
+                 "|",
+                 "(",
+                 ")":
```
thdxg: Could we align this set with ghostty's [Ghostty.Shell.swift](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Ghostty/Ghostty.Shell.swift)? The current set is missing `; & ! # [ ] < > \t`.

Macterm/Ghostty/GhosttyCallbacks.swift
```diff
@@ -58,11 +58,77 @@ final class GhosttyCallbacks: @unchecked Sendable {
...
+    /// and the libghostty Cmd+V clipboard path.
+    static func readPasteboardText() -> String? {
+        let pb = NSPasteboard.general
+
+        // Finder copies files as NSURL data, not strings.
+        if let urls = pb.readObjects(forClasses: [NSURL.self]) as? [URL] {
+            let paths = urls
+                .filter(\.isFileURL)
+                .map { Self.shellEscape($0.path(percentEncoded: false)) }
+                .filter { !$0.isEmpty }
```
thdxg: Small parity nit with ghostty: its `getOpinionatedStringContents()` maps non-file URLs (web URLs, mailto:, custom app schemes, etc.) through `absoluteString` rather than filtering them out. In practice the plain-text fallback at the end usually catches it, since most providers also write the URL as a string, but it's a one-liner to match ghostty exactly: 
```suggestion
            let paths = urls.map { url in
                url.isFileURL ? Self.shellEscape(url.path(percentEncoded: false)) : url.absoluteString
            }
            .filter { !$0.isEmpty }
```

Macterm/Ghostty/GhosttyCallbacks.swift
```diff
@@ -58,11 +58,77 @@ final class GhosttyCallbacks: @unchecked Sendable {
...
+            if !paths.isEmpty {
+                return paths.joined(separator: " ")
+            }
+        }
+
+        if let items = pb.pasteboardItems {
```
thdxg: Is there a real pasteboard provider this branch is meant to catch? `NSPasteboard.readObjects(forClasses: [NSURL.self])` above already handles the standard NSURL / `public.file-url` / `NSFilenamesPboardType` representations, so I'd expect this loop to be unreachable for Finder and the other common sources. Happy to keep it defensively, but a comment naming the provider it targets would help — otherwise it's a second code path that has to track the first (e.g. when fixing the escape set above).

Ghostty itself only has the one read path — see [getOpinionatedStringContents()](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Helpers/Extensions/NSPasteboard%2BExtension.swift#L41-L50). Might be simpler to drop this branch.

thdxg (changes requested):

thdxg:
Thanks for your contribution! I made a few comments in code.

m0n99:
@thdxg Thanks for your comments! Fixed in the latest push.

thdxg:
@m0n99 Thanks. Looks good to me!

PR #59, opened 2026-06-07
fix: hide titlebar background view in native fullscreen to remove top bar artifact

### LINKED ISSUE #24: White bar visible at the top when in fullscreen
### Description

When switching Macterm to fullscreen, a solid white bar appears along the top edge of the screen. This overlay remains persistent and obscures the top portion of the interface. I have attached a screenshot demonstrating the issue where the white bar is clearly visible against the dark theme background.

### Steps to reproduce

1. Launch Macterm.
2. Enter Fullscreen mode (e.g., via the green traffic light button or keyboard shortcut).
3. Observe the top edge of the screen.

### Expected behavior

The app should occupy the entire screen seamlessly without any graphical artifacts or white bars at the top. 

### Macterm version

v1.11.7

### macOS version

macOS 26.5 (25F71)

### Logs or screenshots

Image

mvanhorn:
## Summary

Native fullscreen no longer shows the white bar at the top of the window, fixing #24. The titlebar background view stayed visible when the window entered fullscreen, drawing an opaque strip above the terminal content. `WindowAppearance.swift` now hides that view on the fullscreen-enter notification and restores it on exit, with the existing appearance handling left untouched for windowed mode.

## Why this matters

hujunfeng's report shows the artifact on every fullscreen session, which for a terminal app means a permanently corrupted top edge in the mode many users live in. The fix follows the same NSWindow notification pattern the file already uses for appearance changes, so behavior in windowed and split-view modes is unchanged.

## Testing

Codex review confirmed the project builds and the existing test suite passes with the change. SwiftLint was unavailable in this environment; CI covers it.

AI was used for assistance.

Fixes #24

### DIFF
--- a/Macterm/Views/MainWindow.swift
+++ b/Macterm/Views/MainWindow.swift
@@ -322,6 +322,24 @@ private struct WindowStyler: NSViewRepresentable {
             swiftuiDelegate?.windowDidResignKey?(notification)
         }
 
+        func window(
+            _ window: NSWindow,
+            willUseFullScreenPresentationOptions proposedOptions: NSApplication.PresentationOptions = []
+        ) -> NSApplication.PresentationOptions {
+            // Auto-hide the toolbar together with the menu bar in native
+            // fullscreen. Without .autoHideToolbar AppKit pins the
+            // NSToolbarFullScreenWindow (the titlebar/toolbar strip) across
+            // the top of the screen for the whole fullscreen session, which
+            // is the persistent top bar from issue #24. Hiding its
+            // NSTitlebarBackgroundView only recolors the strip; the window
+            // itself has to auto-hide. (#24)
+            var options = proposedOptions
+            if let inner = swiftuiDelegate?.window?(window, willUseFullScreenPresentationOptions: proposedOptions) {
+                options = inner
+            }
+            return options.union([.fullScreen, .autoHideMenuBar, .autoHideToolbar])
+        }
+
         func windowDidEnterFullScreen(_ notification: Notification) {
             guard let window = notification.object as? NSWindow else { return }
             WindowAppearance.sync(window: window)
--- a/Macterm/Views/WindowAppearance.swift
+++ b/Macterm/Views/WindowAppearance.swift
@@ -226,8 +226,10 @@ enum WindowAppearance {
         // Override the titlebar's private background layer so its color
         // matches the terminal background (or stays transparent when the
         // window is). Without this the titlebar paints its own material
-        // and you get a visible seam at y=titlebarHeight.
-        syncTitlebar(window: window, isTransparent: effectiveTransparent)
+        // and you get a visible seam at y=titlebarHeight. Native fullscreen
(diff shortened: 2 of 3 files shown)


thdxg:
@mvanhorn .Thanks for your contribution! I ran the app on your branch, but the issue seems to persist. Does it work on your machine?

mvanhorn:
You were right - I rebuilt this branch and reproduced it on my machine too: the bar was still there. Dug into the live view hierarchy and found the real cause: the NSTitlebarBackgroundView hiding was landing, but the bar is AppKit's whole NSToolbarFullScreenWindow staying pinned because the window never opts into .autoHideToolbar - recoloring its background can't fix that. f23baa6 replaces the approach: the window delegate now implements window(_:willUseFullScreenPresentationOptions:) returning [.fullScreen, .autoHideMenuBar, .autoHideToolbar], so the toolbar strip auto-hides with the menu bar and only appears on hover (where the existing titlebar sync keeps it matching the theme). Verified by building and running: entering fullscreen is now clean on my machine. Would appreciate a re-test on yours.

thdxg:
@mvanhorn Thanks for taking another look at it! The titlebar now autohides in fullscreen as you said. However, I'm not sure if this would be a correct solution to titlebar color mismatching the background.

On your version, when titlebar hidden:
Screenshot 2026-07-04 at 1 46 35 PM

When titlebar revealed:
Screenshot 2026-07-04 at 1 48 58 PM
Titlebar and the traffic light buttons overlap with content, which causes another visual disturbance.

I think the ideal solution to issue #24 would be painting the titlebar to same color as the background, and any other workaround seems not worth the UX change. And honestly, titlebar color mismatch doesn't look as bad on macOS 27 with their new sidebar design:

Screenshot 2026-07-04 at 1 46 24 PM

I'm grateful for the work you have done, but I think the real solution would be to paint custom color on fullscreen titlebar, which as of now doesn't seem possible.

mvanhorn:
Good catch — you're right that hiding the background view is only half of it. The overlap you're seeing when the titlebar is revealed is the separate, well-known side effect of a transparent full-size-content-view titlebar: the terminal content extends up under the traffic lights and title, so anything in that top strip collides with them. Hiding the background view kills the white bar in fullscreen but doesn't reserve space for the buttons in windowed mode. The usual fix is to inset the content (or top safe-area) by the titlebar height while the titlebar is visible, and drop that inset in fullscreen where it's hidden. Want me to take that on — reserve the traffic-light strip in windowed mode — as part of this or a follow-up? Happy to redo it whichever way you prefer.

mvanhorn:
Agreed - autohiding masks the mismatch rather than fixing it: when the titlebar reveals, its material still doesn't match the terminal background. I think the right fix is to stop relying on the default titlebar material entirely: set `titlebarAppearsTransparent` and drive `window.backgroundColor` from the active terminal profile's background, so the revealed titlebar blends with the content in both light and dark themes. If you're open to it I'll rework the branch that way.

PR #90, opened 2026-06-17
Add sidebar indicators for terminal status

onliner10:
## What

Adds live status indicators to sidebar terminal tabs: spinner while running, green checkmark when a background command completes, and empty circle when idle/active/acknowledged.

## Why

When running multiple terminal-based agents in parallel, it should be easy to see which tabs are still working,which finished and need attention, and which are just idle leftovers.

## How

Adds a small execution-state model on panes, aggregates it at the tab level, and renders that state in the sidebar instead of the static tab icon.

The sidebar treats this as an attention indicator:
- running always shows a spinner
- completed background tabs show a checkmark
- the active tab suppresses completed state and shows idle unless something is currently running
- selecting/interacting with a completed tab acknowledges it back to idle

Command completion is driven by Ghostty command-finished events and the existing foreground-process refresh fallback.

## Verified

Also ran:

- `mise run build --verbose`

## Notes for reviewers

Screenshots attached below show the sidebar states: running spinner, completed checkmark, and idle circle.

https://github.com/user-attachments/assets/047fe65a-2cfa-44e7-8fab-67b432e3177c

### DIFF
--- a/Macterm/App/AppState.swift
+++ b/Macterm/App/AppState.swift
@@ -116,6 +116,7 @@ final class AppState {
             for tab in ws.tabs {
                 for pane in tab.splitRoot.allPanes() {
                     pane.refreshForegroundProcess()
+                    pane.settleTerminalActivityIfQuiet()
                     acknowledgeFinishedCommandIfActive(paneID: pane.id, projectID: projectID)
                 }
             }
--- a/Macterm/App/AppState.swift
+++ b/Macterm/App/AppState.swift
@@ -715,10 +715,15 @@ final class AppState {
     }
 
     func acknowledgeFinishedCommandIfActive(paneID: UUID, projectID: UUID) {
+        // The sidebar shows the *entire* active tab as idle (displayState masks
+        // `.done` for the tab the user is looking at), so every pane in that tab
+        // must actually be cleared — not just the focused one. Otherwise a
+        // non-focused split pane that finished a command stays `.done` under the
+        // hood, gets persisted on quit, and reappears as a checkmark after
+        // restart even though the user saw an empty circle.
         guard NSApp.isActive,
               projectID == activeProjectID,
               let tab = workspaces[projectID]?.activeTab,
-              tab.focusedPaneID == paneID,
               let pane = tab.splitRoot.findPane(id: paneID)
         else { return }
         pane.acknowledgeCommandCompletion()
--- a/Macterm/App/AppState.swift
+++ b/Macterm/App/AppState.swift
@@ -116,7 +116,7 @@ final class AppState {
         // and the quiet-settle only matter when the status indicator is shown;
         // skip them in icon mode so the default poll stays as cheap as before
         // this feature.
-        let trackExecution = Preferences.shared.tabIndicatorMode == .status
+        let trackExecution = Preferences.shared.showTabStatusIndicator
         var didAcknowledgeCompletion = false
         for (projectID, ws) in workspaces {
             for tab in ws.tabs {
--- a/Macterm/App/AppState.swift
+++ b/Macterm/App/AppState.swift
@@ -117,6 +117,7 @@ final class AppState {
         // skip them in icon mode so the default poll stays as cheap as before
         // this feature.
         let trackExecution = Preferences.shared.tabIndicatorMode == .status
+        var didAcknowledgeCompletion = false
         for (projectID, ws) in workspaces {
             for tab in ws.tabs {
                 for pane in tab.splitRoot.allPanes() {
@@ -124,10 +125,15 @@ final class AppState {
                     if trackExecution {
                         pane.settleTerminalActivityIfQuiet()
                     }
-                    acknowledgeFinishedCommandIfActive(paneID: pane.id, projectID: projectID)
+                    didAcknowledgeCompletion = acknowledgeFinishedCommandIfActive(
+                        paneID: pane.id,
+                        projectID: projectID,
+                        saveImmediately: false
+                    ) || didAcknowledgeCompletion
                 }
             }
         }
+        if didAcknowledgeCompletion { saveWorkspaces() }
     }
 
     private func recordProjectVisit(_ projectID: UUID) {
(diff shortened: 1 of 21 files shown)


thdxg:
@onliner10 Thanks for your contribution! I think this is a neat idea but not sure if replacing the original tab icons would be the best user experience. For example, many CLIs are meant to run long-term, not a one-off command, and showing a spinner the whole time might not be what every user wants.

Screenshot 2026-06-19 at 1 27 22 PM

Maybe we can add it as an additional option to the tab icons that users can opt into? Lmk what you think

onliner10:
Thanks for taking a look, and good point - I agree that replacing the tab icon by default is too opinionated, especially for long-running CLIs.

I updated the branch so the default behavior stays the same as today: tabs still use the existing configurable icon, defaulting to `terminal`.

The new status indicators are now an opt-in mode under Settings → Appearance → Sidebar → Tab indicator:

- Icon: existing configurable tab icon behavior
- Status: replaces the tab icon with the running/completed/idle indicator

That way users who prefer stable tab icons can keep them, while folks running multiple parallel one-off agents can opt into the status indicators.

I also re-ran the checks after the update:

- `mise run format --verbose`
- `mise run lint --verbose`
- `mise run test --verbose`
- `mise run build --verbose`

onliner10:
I think I just better understood the main concern here: using the foreground process as the "running" signal is too coarse.

For long-lived interactive CLIs like Claude Code, pi, ssh, etc., the foreground process can stay alive indefinitely, so treating "non-shell foreground process" as running would keep the spinner visible forever. That's not the behavior I'm aiming for.

I'll rework this so the status indicator is based on terminal activity instead: show the spinner while the terminal is actively producing output, then switch to completed/attention-needed once it goes quiet in the background. That should make the feature work much better for tools like Claude Code or pi.

onliner10:
I think I got it to work with long-running/interactive processes like Claude Code, pi, ssh, editors, etc.

The implementation is now a small execution-state tracker that combines several signals instead of relying on a simple “foreground process exists = running” rule.

The tracker listens to:

- Ghostty/shell-integration command-finished events (`COMMAND_FINISHED` / OSC 133;D)
- OSC 9;4 progress reports
- OSC title changes as command-boundary hints, which trigger a fresh foreground-process read
- terminal output/scrollback growth as an activity signal
- periodic foreground process polling
- tty mode inspection via `tcgetattr`

The important part for Claude Code/pi-style tools is the tty mode check. On foreground refresh, Macterm checks whether the foreground process is a shell and whether the pane’s tty is in canonical mode. If canonical input or echo is disabled, we treat the process as raw/cbreak-style interactive. That lets us avoid keeping a tab permanently “running” just because an interactive agent/editor/ssh session remains in the foreground.

So the intended behavior is:

- normal commands like `sleep`, `make`, `npm test`, etc. show as running while foregrounded and flip to done when control returns to the shell
- shell-integration/progress signals can drive explicit `running → done` transitions
- interactive/raw-mode programs do not stay permanently running just because they are foreground
- if an interactive program produces output after user interaction, output-driven activity can still show briefly and then settle after a quiet interval
- restored plain shells ignore startup/shell-integration noise until the user interacts, so reopening the app should not randomly create green checkmarks
- layout `run:` panes still track their initial command because those are intentional startup commands

I also tightened the persistence side: acknowledging a completed tab now saves the cleared state, the active-tab acknowledge path clears the whole tab including split panes, and old workspace snapshots migrate to v4 by clearing stale persisted attention flags.

I rebuilt the app/DMG and tested with my existing workspace state; the random checkmarks no longer reappear for me, and long-running interactive tools behave much better.

Does this match what you had in mind for the status indicator?

thdxg:
@onliner10 Thanks for your work! I tested it, and it feels much more polished now! I have one suggestion. Instead of forcing users to choose between icon or status in the Tab indicator menu, what if we make the icon and the loading spinner coexist?
When a process is running, we can show a spinner, but on idle state, we can show the icon set by the user. So in the settings, we can have a toggle "Show tab loading indicator", which only replaces the icon while it's active.

onliner10:
Done! I implemented your suggestion so the icon and the status indicator coexist instead of being mutually exclusive.

What it does now:

- **Idle**: the user's chosen tab icon shows as-is, nothing extra.
- **Running**: the icon is temporarily replaced by a small spinner (Xcode-build-navigator style) while a command is in the foreground.
- **Done** (a background tab finished a command and awaits attention): the icon stays, with a small solid green status dot in its bottom-trailing corner — like the Messages/FaceTime "available" dot. It's opaque and ringed with the background color so it stays legible over any icon.

The old "Tab indicator: Icon / Status" picker is replaced by a single "Show tab status indicator" toggle, defaulting to off.

I deliberately kept the "done" signal minimal (a dot, not a checkmark badge) because a big green checkmark over the icon read as heavy and not very macOS-native — a small status dot felt closer to how Apple signals a positive/available state on sidebar icons.

How does this look to you? Happy to switch the done indicator to something else (e.g. a checkmark, or nothing at all with just the running spinner) if you'd prefer.

thdxg:
I'm not sure if I'm missing something, but I don't see a green status dot on command success. Could you clarify this?

https://github.com/user-attachments/assets/ce05f792-d7b3-4599-b243-e73e1ff3df0f

onliner10:
Ah, that's by design! The status dot only shows on **background** tabs — the tab you're currently looking at doesn't need an attention signal, since you're already seeing the command finish right in front of you. In your video you're staying in the same tab, so the spinner just stops and the icon returns, with no dot.

To see the dot: enable "Show tab status indicator" in Settings (it's off by default), start a command in one tab, then switch to another tab while it runs. When the background command finishes, the green dot appears on that tab's icon until you visit it again.

I decided not to show the dot on the focused tab because it doesn't require user attention there — the indicator is really meant for tabs you're not currently looking at. Does that distinction make sense to you, or would you prefer the done state to show everywhere?

thdxg:
Thanks for clarifying! That makes perfect sense, and I like the way it is now. I will merge this PR and it will be part of the next release. Thanks again for your contribution!

PR #102, opened 2026-06-21
Make terminal scrolling feel like iTerm2

onliner10:
## What

Routes ordinary terminal scrollback wheel/trackpad events through an iTerm2-style scroll accumulator instead of forwarding every wheel event directly to libghostty. The accumulator converts AppKit wheel/trackpad deltas, including momentum, into whole-row `scroll_to_row` actions while keeping the native overlay scrollbar in sync.

## Why

Macterm scrollback felt sluggish compared with iTerm2 because precise trackpad input was handled by libghostty's row scroll path without iTerm2's fast trackpad accumulation behavior.

## How

Adds a Swift port of iTerm2's default modern scroll accumulator (`fastTrackpad = YES`, sensitivity `1.0`, acceleration `1.0`). `GhosttyTerminalNSView` gives `SurfaceScrollView` first chance to handle ordinary scrollback events when libghostty has not captured the mouse; captured mouse / alternate-screen cases still fall back to libghostty, with momentum metadata preserved.

Scroller drags continue to use `scroll_to_row`, and fractional scroller offsets are floored to whole rows because the terminal core is row-based.

## Verified

## Notes for reviewers

This intentionally stays row-based to match terminal/iTerm2 behavior; it does not attempt pixel-continuous text scrolling.

### DIFF
--- a/Macterm/App/Preferences.swift
+++ b/Macterm/App/Preferences.swift
@@ -61,6 +61,12 @@ final class Preferences {
         didSet { defaults.set(eagerlyStartProjectTabs, forKey: Keys.eagerlyStartProjectTabs) }
     }
 
+    /// Multiplier applied to terminal scroll wheel / trackpad row deltas.
+    /// 1.0 matches iTerm2's default scroll accumulator behavior.
+    var terminalScrollSpeed: Double {
+        didSet { defaults.set(terminalScrollSpeed, forKey: Keys.terminalScrollSpeed) }
+    }
+
     // MARK: - Sidebar icons
 
     var projectIconSymbol: String {
@@ -242,6 +248,7 @@ final class Preferences {
         self.defaults = defaults
         autoTilingEnabled = defaults.bool(forKey: Keys.autoTiling)
         eagerlyStartProjectTabs = (defaults.object(forKey: Keys.eagerlyStartProjectTabs) as? Bool) ?? true
+        terminalScrollSpeed = Self.clampScrollSpeed(defaults.double(forKey: Keys.terminalScrollSpeed), fallback: 1.0)
         windowOpacity = (defaults.object(forKey: Keys.windowOpacity) as? Double) ?? 1.0
         windowBlurRadius = defaults.integer(forKey: Keys.windowBlurRadius)
         windowGlassEnabled = defaults.object(forKey: Keys.windowGlassEnabled) as? Bool ?? false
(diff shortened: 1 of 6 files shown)


thdxg:
@onliner10 Hey, thanks for another contribution! It definitely feels nicer on trackpad. Is it possible to expose a slider in Settings, to let the user configure the scrolling speed?

onliner10:
Done - added a Settings → General → Terminal scroll speed slider . The value is persisted in Preferences and feeds the iTerm-style accumulator as its sensitivity, so 1× preserves the iTerm2 default feel while allowing users to tune it

onliner10:
@thdxg Is that what you meant?

thdxg:
Yes. I will review this soon when I get a chance!

thdxg:
@onliner10 Hi, a few cleanups I'd suggest before merge:

1. **Dead `pow(abs(accumulated), 1)` in `ITermScrollAccumulator.delta`.** `pow(x, 1) == x`, so it's a no-op. Sensitivity is already applied linearly upstream, which is the right choice — so this just reads as "sensitivity lives here" when it doesn't. Suggest dropping it to `Int(accumulated)`.

2. **Mouse-wheel branch passes the unscaled delta to `takeWholePortion`.** In `accumulatedDeltaForMouseWheelEvent` it accumulates `delta * sensitivity` but passes the un-scaled `delta`, while the trackpad branch passes the scaled one. It's harmless today (only the sign is used), but the two paths should match — pass `delta * sensitivity` in both.

3. **The `sensitivity == 1` exact-equality branch splits the algorithm in two.** A `Double` slider can land just shy of 1.0 and silently take the other code path, so the default and "barely moved the slider" behave qualitatively differently. Suggest unifying on the accumulator path and deleting the special-case.

4. **`MainActor.assumeIsolated` in `SurfaceScrollView.deinit`** can trap if the view is ever last-released off the main thread, and it's redundant — `Pane.destroySurface()` already nils `onScrollWheel`, and the closure is `[weak self]`. I'd just drop the line.

5. **Doc nit:** the `scrollWheel` / `SurfaceScrollView` header comments say alt-screen falls through "because mouse isn't captured," but alt-screen is actually handled by the `total > len` check in `canHandleScrollbackWheel` (no scrollback → declines). Worth rewording so a future change to that guard doesn't unknowingly break `less`/`vim` scrolling.

onliner10:
@thdxg All five addressed in 5bec192:

1. Dropped the no-op `pow(abs(accumulated), 1)` → `Int(accumulatedDelta(...))`.
2. Mouse-wheel branch now passes the scaled delta to `takeWholePortion`, matching the trackpad branch.
3. Deleted the `sensitivity == 1` special case — both paths go through the accumulator now.
4. Removed `MainActor.assumeIsolated` from `deinit`.
5. Reworded the `onScrollWheel` / `SurfaceScrollView` docs, and added an inline note on `canHandleScrollbackWheel`, so alt-screen is attributed to the `total > len` guard rather than mouse-capture.

format / lint / test all green.

thdxg:
Looks good!

PR #108, opened 2026-06-22
Fix activity detection for in-place terminal output

onliner10:
## What

Fix tab status spinner/activity detection for tools that repaint the same terminal line instead of producing scrollback.

## Why

I noticed this while running the pi coding-agent harness: its spinner could run without keeping Macterm’s tab spinner active. The same pattern likely affects other tools with in-place progress output, such as Claude Code, Codex, and similar CLIs.

The previous activity signal depended on scrollback/scrollbar changes, so output using carriage returns could repaint the terminal without being detected as ongoing activity.

## How

Use libghostty render actions as an activity heartbeat for commands that are already considered running. Render events are not allowed to start the spinner by themselves, because renders also happen for prompt redraws and input echo.

Also treat shell interpreters running scripts or `-c` commands as foreground work rather than idle shells, so shebang scripts are classified correctly.

Finally, simplify execution tracking around explicit running sources: foreground, activity, and progress.

## Verified

## Notes for reviewers

Manual repro script used for verification:

```bash
#!/bin/bash
spinner=("⠋" "⠙" "⠹" "⠸" "⠼" "⠴" "⠦" "⠧" "⠇" "⠏")
end_time=$(($(date +%s) + 10))

while [ "$(date +%s)" -lt "$end_time" ]; do
  for s in "${spinner[@]}"; do
    echo -ne "\r$s Testing."
    sleep 0.15
    if [ "$(date +%s)" -ge "$end_time" ]; then
      break
    fi
  done
done

echo -ne "\rDone!          \n"
```

### DIFF
--- a/Macterm/Ghostty/GhosttyCallbacks.swift
+++ b/Macterm/Ghostty/GhosttyCallbacks.swift
@@ -68,6 +68,12 @@ final class GhosttyCallbacks: @unchecked Sendable {
             let s = action.action.scrollbar
             DispatchQueue.main.async { view.surfaceDidUpdateScrollbar(total: s.total, offset: s.offset, len: s.len) }
             return true
+        case GHOSTTY_ACTION_RENDER:
+            guard let view = surfaceView(from: target) else { return false }
+            DispatchQueue.main.async { view.surfaceDidRender() }
+            // Do not consume the action: keep libghostty's existing render path
+            // unchanged, and use this only as an activity signal.
+            return false
         case GHOSTTY_ACTION_RELOAD_CONFIG:
             // libghostty fires this (with soft = true) when a surface's
             // conditional state changes — notably on set_color_scheme, which
--- a/Macterm/Model/SplitNode.swift
+++ b/Macterm/Model/SplitNode.swift
@@ -64,31 +64,31 @@ private struct ForegroundProcessKey: Equatable {
     }
 }
 
+private enum TerminalExecutionSource: Equatable {
+    case foreground
+    case activity(Date)
+    case progress
+}
+
 private struct TerminalExecutionTracker {
     init(hasUserInteraction: Bool = false) {
         self.hasUserInteraction = hasUserInteraction
     }
 
-    /// The foreground process seen on the last poll (nil = shell / none).
-    /// Transitions are driven by *changes* to this, not by re-deriving state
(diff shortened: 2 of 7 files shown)


onliner10:
@thdxg could you take a look at this? It'd really help with my daily work, to have Macterm correctly detect activity

thdxg:
@onliner10 Thanks for your patience 🙂 I think this pr is a really nice catch. I ran your script on the current release and your pr version, and confrimed the expected behaviors. Could you address one thing:

Add a tracker-level test for the core invariant. The "a render can keep a spinner alive but never start one" rule lives only in the onTerminalRender closure, which isn't covered. Since the tracker is unit-testable, please pin:
- markTerminalActivity from .idle with no prior interaction stays .idle
- markTerminalActivity from .done does not return to .running
- an activity-sourced run still quiet-settles after the interval

Otherwise a future refactor could silently reintroduce the "prompt redraw keeps spinning" bug with nothing to catch it.

onliner10:
Thanks! Done — the tracker is now directly unit-tested. I promoted `TerminalExecutionTracker` to internal and added `TerminalExecutionTrackerTests` pinning all three cases, plus a couple of boundary tests:

- `markTerminalActivity` from `.idle` with no prior interaction → stays `.idle`
- `markTerminalActivity` from `.done` → does **not** return to `.running`
- an activity-sourced run still quiet-settles after the interval
- (boundary) `.idle` *with* interaction → starts a run, and a heartbeat while `.running` refreshes the settle timestamp

One thing the `.done` test surfaced: that case wasn't actually holding. With prior interaction, `markTerminalActivity` fell through the `hasUserInteraction` guard and returned `.running`, so output after a command finished (e.g. a background job) could resurrect the checkmark — the "never start" rule really did live only in the `onTerminalRender` closure, as you noted. I added an early `guard currentState != .done`, so a render/output heartbeat can keep a run alive but can never flip `.done` back to `.running`; only a new foreground process or an explicit progress marker can.

That's a small behavior change (output while `.done` no longer restarts the spinner), which I think is the more correct read of the invariant — but flagging it in case you'd rather keep the old resurrection path.

All tests pass (`mise run test`), plus format/lint.

thdxg:
@onliner10 Thanks for your work!

PR #109, opened 2026-06-24
Paste clipboard images into TUIs instead of dropping them

OpenSource03:
## What

Pasting a raw image from the clipboard (a screenshot, "Copy Image", etc.) into the terminal now writes the image to a temp PNG and pastes its file path, instead of inserting nothing.

## Why

Pasting an image into a TUI that reads images from a pasted path — e.g. Claude Code — does nothing in Macterm, even though it works in base Ghostty, Terminal.app, and others. Macterm overrides libghostty's clipboard-read callback with a text-only resolver (`readPasteboardText`), so an image-only pasteboard returns an empty string and the paste is dropped on the floor.

## How

`readPasteboardText()` gains an image fallback: when the pasteboard has no text/file-URL but does carry `.png`/`.tiff` (or an `NSImage`), the image is normalized to PNG, written to a temp dir, and its shell-escaped path is returned — matching base Ghostty's behavior and what path-reading TUIs consume. Text and Finder file-URLs still take precedence, so normal pastes are unchanged.

A side-effect-free `hasPasteboardContent()` (a cheap `availableType(from:)` check) now drives the context-menu Paste item's enablement, so opening the menu no longer writes a temp PNG just to test whether anything is pasteable.

The resolvers take an injectable `NSPasteboard` (defaulting to `.general`) so the behavior is covered by unit tests that drive a real, isolated pasteboard with a real PNG.

## Verified

### DIFF
--- a/Macterm/Ghostty/GhosttyCallbacks.swift
+++ b/Macterm/Ghostty/GhosttyCallbacks.swift
@@ -131,9 +131,12 @@ final class GhosttyCallbacks: @unchecked Sendable {
     /// Returns pasted text from the pasteboard: file paths (Finder drag/copy)
     /// fall back to plain string. Called by both the context-menu paste path
     /// and the libghostty Cmd+V clipboard path.
-    static func readPasteboardText() -> String? {
-        let pb = NSPasteboard.general
-
+    ///
+    /// When the pasteboard holds a raw image (e.g. a screenshot) and no text,
+    /// the image is written to a temporary PNG file and its path is returned —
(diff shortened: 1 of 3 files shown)


thdxg:
@OpenSource03 Thanks for your contribution! This looks mostly good to me. I'd be happy to merge this once you resolve the one comment!

OpenSource03:
> @OpenSource03 Thanks for your contribution! This looks mostly good to me. I'd be happy to merge this once you resolve the one comment!

Thanks! Done.

thdxg:
@OpenSource03 Thanks for your work!

PR #119, opened 2026-06-30
fix: render CJK IME preedit text

### LINKED ISSUE #118: Japanese IME preedit text is not displayed
### Description

Japanese IME preedit text is not displayed while composing text.

When using a Japanese IME, the uncommitted/pre-conversion text does not show up in the terminal while typing. The committed text appears after conversion/commit, but the in-progress composition is
  invisible.

  - Native Ghostty does not seem to have this problem.
  - This appears to affect non-ASCII preedit text such as Japanese kana/kanji.

While looking through the source, this may be related to `GhosttyTerminalNSView.setMarkedText` passing `String.count` to `ghostty_surface_preedit`:

  ```swift
  text.withCString { ghostty_surface_preedit(surface, $0, UInt(text.count)) }
  ```

`ghostty_surface_preedit` appears to expect a UTF-8 byte length. For Japanese text, `String.count` is shorter than the UTF-8 byte count, so the preedit string may be truncated before libghostty decodes/
  renders it.

### Steps to reproduce

1. Open Macterm.
2. Switch the macOS input source to a Japanese IME.
3. Type Japanese text, for example `かな`, without committing/converting yet.
4. Observe the terminal while the IME composition is still active.

### Expected behavior

The preedit/pre-conversion text should be visible at the cursor while composing.

### Macterm version

Version 1.18.2 (1.18.2), GitCommit `9405191`

### macOS version

macOS 26.3.1 (25D771280a) arm64

### Logs or screenshots

Possible fix: pass the UTF-8 byte length to `ghostty_surface_preedit`, for example `UInt(text.utf8.count)` or `UInt(text.utf8CString.count - 1)`, instead of `UInt(text.count)`.

macterm: 

https://github.com/user-attachments/assets/80e8465a-7cce-4d81-aae4-0f883ce43f5b

ghostty:

https://github.com/user-attachments/assets/f7a489bb-84fe-4841-a7bc-f93b2d4e22d0

takuma-ru:
## What

Fix CJK IME preedit text rendering while composing.

Macterm now passes the UTF-8 byte length of marked text to `ghostty_surface_preedit`, instead of Swift's character count. It also reports the marked range length with `String.UTF16View`, matching `NSRange`/AppKit expectations.

## Why

CJK IME preedit text can be invisible while composing. I reproduced this with Japanese IME input: the committed text appeared after conversion, but the in-progress kana/kanji composition did not render at the cursor.

`ghostty_surface_preedit` takes a UTF-8 pointer and byte length. For CJK/non-ASCII text, `String.count` is shorter than the UTF-8 byte count (`"あ"` is 1 Swift character but 3 UTF-8 bytes), so libghostty could receive a truncated UTF-8 sequence and fail to decode/render the preedit text.

Closes #118

## How

`GhosttyTerminalNSView.setMarkedText` now computes the preedit length from `text.utf8.count` before calling `ghostty_surface_preedit`.

The marked range length is computed with `text.utf16.count`, so AppKit-facing range values stay in UTF-16 code units.

No new tests are added; the change is a narrow AppKit/libghostty bridge fix that uses Swift's standard UTF-8 and UTF-16 views directly at the call site.

## Verified

- [ ] (no model / persistence / palette / hotkey logic changed) Added or updated tests for new model / persistence / palette / hotkey logic

## Notes for reviewers

Verified Japanese IME preedit text is visible while composing in the built app.

Recordings:

- Before: Japanese IME preedit text is invisible while composing.
  
  https://github.com/user-attachments/assets/c49b10c3-1d98-4b59-a9df-ef59399cf1bd
- After: Japanese IME preedit text is visible at the cursor before commit/conversion.
  
  https://github.com/user-attachments/assets/5a06d8f0-f723-4326-8dcc-15e2e5db3884

Also ran:

- `mise run format --verbose`
- `mise run test --verbose`
- `mise run build --verbose`

### DIFF
--- a/Macterm/Views/Terminal/GhosttyTerminalNSView.swift
+++ b/Macterm/Views/Terminal/GhosttyTerminalNSView.swift
@@ -913,9 +913,9 @@ extension GhosttyTerminalNSView: @preconcurrency NSTextInputClient {
     func setMarkedText(_ string: Any, selectedRange: NSRange, replacementRange: NSRange) {
         guard let surface else { return }
         let text = (string as? String) ?? (string as? NSAttributedString)?.string ?? ""
-        _markedRange = text.isEmpty ? NSRange(location: NSNotFound, length: 0) : NSRange(location: 0, length: text.count)
+        _markedRange = text.isEmpty ? NSRange(location: NSNotFound, length: 0) : NSRange(location: 0, length: text.utf16.count)
         _selectedRange = selectedRange
-        text.withCString { ghostty_surface_preedit(surface, $0, UInt(text.count)) }
+        text.withCString { ghostty_surface_preedit(surface, $0, UInt(text.utf8.count)) }
     }
 
     func unmarkText() {


thdxg:
@takuma-ru Thanks for your contribution! This PR looks good to me!

PR #120, opened 2026-06-30
Useability KeyBindings

### LINKED ISSUE #113: Some MacTerm UI Improvements
### Description

Thanks for this great tool!

### Motivation

### Alternatives considered

_No response_

aryeh:
## What

Add Key Bindings to
- [x] Rearrange tab higher in project
- [x] Rearrange tab lower in project
- [x] Select previous tab in project
- [x] Select next tab in project 

- [x] Focus and Open SideBar temporarily  until new tab is selected, or ESC defocuses SideBar
- [x] Allow navigating SideBar with arrow keys
- [x] Prevent active tab from grabbing focus when a tab in the SideBar is selected/highlighted but not formally chosen

## Why

Allow for keyboard driven reorganizing and managing tabs.

Closes #113 

## Verified

- [x] Checked by testing app


thdxg:
@aryeh Thanks for your contribution! The test CI seems to be failing. Could you resolve the issue?
Also, it seems like your changes for PR #115 is included in this PR as well. Could you narrow the scope of this PR to just include keybinds?

aryeh:
will fix and submit a new request.

PR #122, opened 2026-07-03
fix: correct off-by-one in search match counter

jcuhnoio:
## Summary
- Terminal search's match counter rendered ghostty's zero-based `selected` index directly, so the first match showed "0 of 5" and the counter topped out at "4 of 5" instead of "5 of 5".
- `TerminalSearchState.displayText` now adds 1 when formatting the selected index.

## Test plan
- [x] Added `TerminalSearchStateTests` covering `displayText` at first/last match and the no-selection/no-total fallbacks.
- [x] `mise run test --verbose`


jcuhnoio:
Before:
image

After:
image

PR #152, opened 2026-07-06
Track the system appearance instead of reading back our own pinned one

### LINKED ISSUE #144: System Appearance compliance
### Description

Macterm does subscribe to system appearance change events. When toggling system appearance from light to dark, Macterm does update, toggling back, Macterm keeps using dark appearance.

### Steps to reproduce

1. Make sure to have themes correctly set up in Ghostty conf, e.g. `theme = dark:Melange Dark,light:Melange Light` (theme exists in `/.../Macterm.app/Contents/Resources/ghostty/themes`)
2. Launch Macterm 
3. Check current system appearance is picked correctly
4. Toggle system appearance: Macterm toggles appearance
5. Toggle system appearance: Macterm does not toggle appearance

### Expected behavior

Macterm respects system appearance

### Macterm version

1.19.0 (9948f04)

### macOS version

macOS 27.0 beta

### Logs or screenshots

_No response_

jcuhnoio:
Fixes #144.

## Problem

With a `theme = light:X,dark:Y` split, toggling the system appearance flipped Macterm once — and then it latched. Dark → light never came back.

## Cause

A feedback loop. `AppColorScheme` applies `.preferredColorScheme` derived from the *resolved theme*, which holds the window appearance (and on newer macOS, the app's). But every input to theme resolution read that same appearance back:

- the change detector was a KVO on `NSApp.effectiveAppearance`, which goes silent once the appearance is pinned;
- chrome split resolution (`currentScheme`) read `NSApp.effectiveAppearance`;
- surfaces re-derived their split side only in `viewDidChangeEffectiveAppearance`, which a window held by our own `preferredColorScheme` never delivers.

After one switch, every input echoed our own output and nothing re-resolved.

## Fix

Break the cycle at both ends:

- **Detection**: `GhosttyApp` now tracks the OS-level scheme. The decision logic is pure and tested (`ThemeResolver.systemScheme`): trust `effectiveAppearance` only while no app-level override is pinned (this also tracks "Auto" mode precisely); otherwise read the global-domain `AppleInterfaceStyle` default via CFPreferences, which nothing the app pins can shadow. A `DistributedNotificationCenter` observer for `AppleInterfaceThemeChangedNotification` joins the existing KVO — the OS posts it on every toggle regardless of pinning. Handling is deduped on the resolved scheme since an unpinned switch fires both.
- **Application**: surfaces key `ghostty_surface_set_color_scheme` off the tracked system scheme (creation + `viewDidChangeEffectiveAppearance` re-assert), and on a system switch the scheme is pushed to every live surface directly. Each push soft-reloads config and re-emits `CONFIG_CHANGE`, refreshing the chrome's `resolvedColors` with the new side; the stale snapshot is dropped up front so color accessors fall back to the theme file for the new scheme in the interim (also covers the zero-surface case).

A `system appearance changed: dark|light` log line (subsystem `GhosttyApp`) makes future diagnosis of this class of bug one `mise run logs` away.

## Testing

- New unit tests for the pure scheme decision (`ThemeResolverTests`), including the exact #144 latch case (app pinned dark, system back to light, global default absent → light).
- `mise run format` / `lint` / `test` all pass (one unrelated flaky failure in `AppStateTests.closePane_kills_only_that_panes_session` on a single run; passes on re-run and is untouched by this change).
- Not yet verified live on macOS 27 beta (the reporter's environment)

### DIFF
--- a/Macterm/Ghostty/GhosttyApp.swift
+++ b/Macterm/Ghostty/GhosttyApp.swift
@@ -22,6 +22,12 @@ final class GhosttyApp {
     private var resourcesDir: String?
     @ObservationIgnored
     private var appearanceObserver: NSKeyValueObservation?
+    @ObservationIgnored
+    private var systemAppearanceObserver: (any NSObjectProtocol)?
+    /// The OS-level light/dark scheme. Split resolution must key off this —
+    /// never an `effectiveAppearance`, which our own `.preferredColorScheme`
+    /// pins, latching the theme after one system switch (issue #144).
+    private(set) var systemScheme: ThemeResolver.Scheme = .light
     /// Chrome colors as libghostty resolved them for a live surface — the
     /// active `theme = light:X,dark:Y` side already applied. Populated from
     /// `GHOSTTY_ACTION_CONFIG_CHANGE` (see `adoptResolvedColors`) and preferred
@@ -31,6 +37,7 @@ final class GhosttyApp {
     private var resolvedColors: ResolvedColors?
 
     private init() {
+        systemScheme = Self.readSystemScheme()
         resolveResources()
         guard ghostty_init(UInt(CommandLine.argc), CommandLine.unsafeArgv) == GHOSTTY_SUCCESS else {
             logger.error("ghostty_init failed")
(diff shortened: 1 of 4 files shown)


jcuhnoio:
https://github.com/user-attachments/assets/f5355c70-4846-41cb-8491-86fe13f6d03d
Left is current, right is fixed
(Excuse the recording, I was too lazy to screen record 🤫)

thdxg:
@jcuhnoio Thanks!! Looks good to me

PR #158, opened 2026-07-07
Audit remediation + sidebar tab/project drag-and-drop, OSC 2 title fixes

thdxg:
## Overview

This branch started as codebase-audit remediation and grew two adjacent pieces of work. Since it squash-merges, the scope is spelled out here.

### 1. Audit remediation (the original scope)
Data-loss, race, perf, and convention fixes across the codebase:
- **Data-loss:** remote panes no longer lose their remoteness on restart (`projectPath` is identity/verbatim, `workingDirectory` a local-only hint, gated on `isRemote` in snapshot/split/serializer); a corrupt persistence file is no longer clobbered by the next autosave (both `WorkspaceStore` and `ProjectStore` latch `loadFailed` and refuse to save); control-socket accepted fds clear inherited `O_NONBLOCK` (no more spurious `bad_request`/truncation).
- **Security:** the `~$(cmd)` command-injection hole in `RemoteSpawn` is closed (strict username allowlist), and the single-quote-free wire-format invariant is now runtime-enforced pre-`shellQuote`.
- **Perf:** cached per-keystroke hotkey parsing, poll timer only rebuilds on a cadence change, theme colors/CLI probe cached, no subprocess spawn in a SwiftUI `body`.
- **Conventions:** `.public` on Logger interpolations, `@Observable` throughout (incl. `Updater`), `MactermTheme` for colors, `FocusRestoration` for reshape-adjacent focus, tempdir-injected tests, plus CI/script hardening (cache stale-loop, release-tag hard-fail, nullglob, per-clone git identity).

### 2. Sidebar tab/project drag-and-drop (new feature)
Drag a tab to another project (or reorder within one), and drag project headers to reorder — native `Transferable` payloads (`MovableTab`/`MovableProject`), exported UTTypes, `adoptTab(at:)`/`moveTab(toIndex:)`/`rebind(projectID:)`. Replaces the List `.onMove` (which hijacked row drags). The right-click "Move to Project" menu remains as a fallback.

### 3. OSC 2 title + hotkey fixes
- A program's OSC 2 title (e.g. `claude`'s) is kept instead of snapping back to the process name; `zmx`/version-string titles are resolved to the real program name.
- Repaired hotkeys that an earlier commit on this branch had broken by stamping the SwiftUI window's identifier (fixed per the documented "never stamp the WindowGroup window" rule).

## Review follow-ups addressed
- **High:** `SO_SNDTIMEO` on the accepted control-socket fd (the `O_NONBLOCK` fix otherwise traded truncation for a main-actor hang).
- Regression tests for all three headline data-loss fixes; OSC 52 confirm-prompt flood cap + content preview; `ZmxClient` drain bounded (no forever-block); control-connection concurrency cap; op-path quote-violation exits non-zero.

## Testing
`mise run format`, `lint`, `test` all green. Drag-and-drop gesture verified manually (can't be scripted headlessly).


thdxg:
On the `benchmark:regression` flag (`workload-focused` CPU +39%, CPU ms/s +40%): this looks like **shared-runner noise**, not a code regression, and the numbers are internally inconsistent with a real one.

The three `workload-*` states run an **identical** load (same busy tabs, same date-loop per pane) — they differ only in window state. A genuine per-pane/per-poll CPU regression would raise CPU across **all three** workload states together. Instead:

- `workload-focused` CPU: **+39%** 🔺
- `workload-unfocused` CPU: **−4%**
- `workload-minimized` CPU: **−11%**
- RSS: **−3% to −4%** across all workload states

Focused regressing while the identically-loaded unfocused/minimized states *improve*, and memory dropping everywhere, is the signature of a busier runner on the focused sample — exactly the cross-run variance the report footnote warns about (single 30s sample, different shared runners).

I also traced the hot paths this branch touches and none add per-tick work — the KERN_PROCARGS2 argv parse now stops after `argc` fields instead of decoding the whole environment block (cheaper), and the OSC-title replay guard removes redundant per-rewire sysctl/poll churn.

That said, I folded in the one poll-path optimization from the audit that hadn't landed (`reschedulePoll` was rebuilding the `Timer` on every `.terminalPollEvent` even when the cadence was unchanged — frequent under a busy workload). It's a genuine reduction in per-event churn on the workload path and should help the metric settle, at zero behavior change (cadence values untouched; `PollCadenceTests` still green). Pushed in 87e6d24.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

thdxg:
Thanks for the thorough review — all findings addressed in b445a1e (title/body also updated to reflect the real scope).

**High #1 — O_NONBLOCK fix trades truncation for a main-actor hang.** Fixed. Set `SO_SNDTIMEO` alongside `SO_RCVTIMEO` on the accepted fd, so a client that stops reading can't block `write(2)` on the main actor past the timeout — the write loop's EAGAIN branch terminates the (truncated) send and logs it. You were right that this was the more dangerous failure mode than the bug being fixed.

**Medium #2 — no regression tests for the headline data-loss fixes.** Added: `snapshot_persists_remote_spec_verbatim_with_nil_working_directory` + `remote_pane_round_trips_as_remote`, and `save_refuses_after_corrupt_load_...` for **both** `WorkspaceStore` and `ProjectStore` (write garbage bytes, assert the store loads empty and the next save leaves the file byte-identical). These guard exactly the regressions the PR exists to prevent.

**Medium #3 — OSC 52 prompts don't coalesce.** Fixed: a `@MainActor` in-flight guard shows one alert at a time and drops the rest (a looping TUI can't stack modals), and the prompt now previews the (flattened, truncated) content so the user can judge it.

**Medium #4 — zmx drain can block a pool thread forever.** Fixed: replaced the blocking `availableData` drain with an await on the readabilityHandler's own empty-chunk EOF signal, bounded at 1s — so a descendant holding the stdout write end can't wedge the thread. Returns whatever buffered on timeout.

**Medium #5 — unbounded connection queue.** Fixed: a `DispatchSemaphore(value: 16)` caps concurrent in-flight connections; over-cap connections are refused fast (closed immediately) rather than parking a worker behind the stalled ones.

**Low — op-path quote diagnostic exits 0.** Fixed: `assertSingleQuoteFree` now takes a fallback mode — the op path (kill/probe) exits non-zero (honest failure), the pane path still drops to a shell (a bare exit would vanish the pane).

**Low — "close-race removed entirely" comment.** Softened to note `stop()`'s 1s fallback narrows the race to a wedged-accept-thread window rather than eliminating it.

**Deferred (deliberately):** the silent save-refusal *surfacing* (a user-facing warning) — the refusal itself is the correct default and is now tested, but wiring a new alert would grow this already-over-scoped PR; I'll do it as a focused follow-up. The over-cap "empty or unreadable request" message is left generic (the real reason is logged) for the same reason.

Not changed: the empty-`argv[0]` KERN_PROCARGS2 ambiguity (inherent to the format, as you noted) and the Settings live-update tradeoff (justified by the `UserDefaults.standard` ban).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #164, opened 2026-07-14
Make the window-state benchmark tolerant of shared-runner CPU noise

thdxg:
## What

Two changes to `scripts/benchmark.py` so the window-state benchmark stops flagging shared-runner CPU jitter as regressions:

1. **Median-of-K sampling** — each state is observed as the median of `--samples` short windows (default 3×10s) instead of one 30s window. Same total wall-clock; a single co-scheduled CPU spike lands in one window and the median discards it.
2. **Baseline gate** — a CPU delta is no longer flagged when the *baseline itself* is noise-dominated (`cpu_pct` < 1.5, `cpu_ms/s` < 15). A big relative swing off a 0.5% baseline is jitter; off a 2% baseline it's real and still flags.

## Why

PR #163 was labeled `benchmark:regression` off pure runner noise: focused-idle CPU read 0.57% on main's runner and 1.43% on the PR's — a +150% "regression" that nothing in that diff (an enum case + a pasteboard write on user action) could cause. The benchmark compared one 30s sample against one sample from a *different* shared runner; at sub-1% CPU the measurement variance dwarfs the 25% threshold, so noise clears the bar routinely.

The two defenses are complementary. Median-of-K attacks the variance at the source (an n=1 vs n=1 estimator is the root problem); the baseline gate is the principled way to stop "big % of a tiny number" without blinding the benchmark to real regressions in states that run at meaningful CPU.

## How

- `sample_state` split into a `sample_window` primitive + a medianing wrapper. `--samples 1` returns exactly one window (old behavior verbatim — local one-shot runs unchanged). Powermetrics fields that go null on virtualized runners are median-of-present, never counted as 0.
- `METRICS` gains a `min_baseline` per metric (CPU only; `None` for RSS/wakeups). Enforced in `significant_pct`, which governs the 🔺/🔻 arrow **and** the label. `delta_cell` still prints the raw % regardless — a gated delta is visible, just not flagged.
- Results JSON bumped to schema 2: `seconds_per_state` stays the *total* observation time (back-compat with a schema-1 reader), plus new `seconds_per_window` / `samples_per_state`. The report footnote reads from these and explains both mechanisms; it degrades gracefully on an old schema-1 baseline.
- `bench.sh` passes `3×10s` defaults, env-overridable via `BENCH_SAMPLES` / `BENCH_SECONDS`. The workflow calls `mise run bench` unchanged, and per-state wall-clock stays ~30s so the 30-min CI timeout is safe.

## Verified

- [x] Unit-tested the gating + median logic against PR #163's exact numbers (14 assertions): the focused/idle rows now show their % but produce no arrow and no label, while a genuine regression off a meaningful baseline (2.0→3.0% CPU, 17.2→22.9 ms/s) still flags. Median tames a synthetic single-window spike (0.6/5.0/0.7 → 0.7).
- [x] Rendered the full report end-to-end from #163's numbers and from a schema-1 baseline — table + footnote correct in both, back-compat holds.
- [ ] **Full app-driven run happens on this PR's CI** — my local desktop denies the app activation the benchmark needs (the documented `open`-based window-creation requirement), so the split-sampling path against a live app is exercised here rather than locally. That's the main thing to watch on this run.

## Notes for reviewers

- The current `main` baseline is a schema-1, single-30s-window result, so this PR's comparison table won't *visibly* show the variance reduction — it compares medianed numbers against old single-window numbers (still valid; both measure the same states). The baseline becomes schema-2-vs-schema-2 once this merges and the next PR compares against it.
- Tuning constants (`--samples 3`, `min_baseline` 1.5 / 15) are deliberate but adjustable — happy to dial either if you'd rather trade a bit more tolerance for sensitivity or vice-versa.
- Followed on from the discussion on #163; the label there was a confirmed false positive.


thdxg:
Clearing the `benchmark:regression` label — it's a false gate for this PR, and the run actually **confirms the fix works**.

**The CPU fix did its job.** Every idle-CPU delta that would have flagged under the old logic is now correctly suppressed:
- `unfocused` CPU ms/s 6.4 → 10.9 (+70%) — shows in the table, **no flag** (baseline 6.4 < 15 gate)
- `focused` / `minimized` CPU rows (+50–69%) — all shown, none flagged
- `workload-unfocused` CPU ms/s 18.8 → 23.6 (+26%) — cleared the relative bar but caught by the 5.0 absolute floor (Δ 4.8)

None of these produced a 🔺 or a label. On `main`'s old single-window + floor-only logic, several would have.

**The label is driven entirely by workload Memory** (+28–32%, ~+36 MB), which:
1. **Cannot be a regression this PR introduced** — the diff is CI-only (`benchmark.py` / `bench.sh` / `AGENTS.md`); the measured app binary is byte-identical to `main@ed41b64`.
2. Is either a pre-existing delta on `main` or workload-memory run-to-run variance — the workload spawns 8 real shells+surfaces, and this compares against an *old single-window* baseline, so a single cross-methodology comparison can't tell the two apart.

Per discussion, merging as-is and **not** adding a workload-memory guard yet — better to let a few schema-2-vs-schema-2 runs establish whether workload RSS is genuinely up or just noisy than to mask a possible real leak. Tracking that separately.

Issue #168, opened 2026-07-14
Guard against fork drift shipping a stale GhosttyKit/zmx


thdxg:
### Motivation

Macterm ships prebuilt GhosttyKit and zmx from the `thdxg/ghostty` and `thdxg/zmx` forks. Both forks' `sync-upstream.yml` had a bug that silently reverted ~190 upstream source files (including a libghostty reflow fix) to stale versions while looking up-to-date — which is how #112 reached users despite being fixed upstream months earlier. Nothing detected the drift; it surfaced only when a user reported the downstream symptom.

The sync workflows are now rewritten to be revert-proof (overlay fork-owned files + apply known in-repo patches, fail loudly if a patch won't apply). But there's still no positive assertion that the shipped artifact is built from current-ish upstream.

### Proposal

A guard that catches fork drift automatically. Options (any/all):

1. **Fork CI assertion** — a step (or scheduled workflow) in each fork that asserts `main` == upstream `main` + the declared downstream allowlist/patches, and fails on any other divergence.
2. **Macterm setup warning** — `scripts/setup.sh` warns (or a CI check fails) when the downloaded GhosttyKit's `include/ghostty.h` / version is older than upstream's current, i.e. the artifact is stale.
3. **Freshness check** — assert the fork's last successful build is from a commit whose upstream base is within N days of upstream HEAD.

### Context

See the fork-sync fixes done in July 2026 (rewritten `sync-upstream.yml` in both forks; backups at `refs/backup/main-pre-sync-fix`). This issue is the "never let it silently happen again" follow-up.

### Area

`area:ci`.

thdxg:
Done. Three parts, all verified:

- **This repo** (#169, merged): `scripts/setup.sh` warns when the bundled GhosttyKit's upstream base is materially behind ghostty-org/ghostty HEAD.
- **thdxg/ghostty** + **thdxg/zmx**: scheduled `assert-no-drift.yml` (thdxg/ghostty#4, thdxg/zmx#2) asserts each fork carries only its declared downstream delta on pristine upstream — anchored on the merge-base with upstream, with a freshness window and an allowlist-of-changed-files check (the real revert-detection). Both now run green on `main` (base == upstream HEAD, 0 behind, all changed paths allowlisted).

The first live run surfaced (and I fixed) a merge-base-vs-HEAD~1 bug: a normal PR merge leaves >1 commit ahead until the next daily sync recollapses it, so the check now anchors on merge-base rather than commit count.

Issue #167, opened 2026-07-14
Debug-only `macterm pane resize` to drive an isolated in-place surface resize


thdxg:
### Motivation

Diagnosing resize/reflow bugs (#112) means answering questions like "is the corruption from the NSView reparent, from repeated `set_size` calls, or from a single large-delta resize?" Today the only way to resize a surface is through the full SwiftUI layout path, so those variables are **confounded** — you can't isolate one. I ended up unable to cleanly A/B them without adding a throwaway Darwin-notification hook.

### Proposal

A **debug-only** affordance to drive `ghostty_surface_set_size` on a pane directly, in-place, once — bypassing SwiftUI layout:

```
macterm pane resize --pane N --cols C --rows R   # debug builds only
```

Gated behind a debug build / `MACTERM_DEBUG` env flag so it never ships in release. Lets you reproduce a specific resize transition (e.g. "132×65 → 63×31 in one step") deterministically and compare against the gradual/divider path.

### Alternative

A `MACTERM_DEBUG` Darwin-notification command (like `BenchmarkControl`) instead of a CLI verb — either works; the point is an isolated, scriptable single resize.

### Area

`area:terminal`, `area:tests`, CLI (`macterm`).

thdxg:
Done in #169 (merged to main): debug-only `macterm pane resize --cols --rows`, gated `#if DEBUG` in both the CLI and ControlHandler dispatch. Closing manually (squash subject dropped the keyword).

Issue #166, opened 2026-07-14
Expose zoom (and split-resize) as control-socket verbs


thdxg:
### Motivation

Zoom is triggered only by a keybind (Cmd+Shift+Enter). There's no control-socket verb for it, so scripting or reproducing zoom-related behavior (like #112) requires `osascript`/System Events synthetic keystrokes — which need an Accessibility grant and are flaky/non-deterministic. The same applies to other UI-only actions (split-divider resize, search).

### Proposal

Expose UI actions that currently have no headless path as control-socket verbs:

```
macterm pane zoom [--pane N]              # toggle zoom on a pane
macterm pane resize-split --pane N --ratio 0.3   # set a split's ratio
```

These call the same `AppState`/`TerminalTab` methods the keybinds do (`toggleZoom(paneID:)`, the branch `ratio` setter), so no new UI logic — just socket plumbing in `ControlHandler`, matching the existing `pane split/focus/close/run` pattern.

### Why

Makes zoom and split-resize deterministically scriptable for both debugging and the integration/e2e test layer (#121). During #112 I had to drive zoom via `osascript` keystrokes, which is exactly the kind of non-reproducible path this removes.

### Area

`area:terminal`, CLI (`macterm`).

thdxg:
Done in #169 (merged to main): `macterm pane zoom` and `pane resize-split --axis --ratio`. The PR's squash subject dropped the closing keyword, so closing manually.

Issue #112, opened 2026-06-26
scrollback breaks after zoom/un-zoom a pane


armyers:
### Description

Zoom a pane that isn't full-width/full-height (cmd-shift-enter), then un-zoom, and the scrollback that was there comes back empty - the screen looks like it was just cleared, with the prompt at the top. New output works fine; only the pre-existing scrollback is gone.

- Only happens when the pane is both less than full height and less than full width
- Dragging the split dividers to resize the same pane (even drastically) does not cause it - only zoom does.
- Native Ghostty terminal does NOT have this problem.
- (edited) also, both cmux and supacode are built using libghostty-vt and they do not have this problem, just for reference :)

### Steps to reproduce

1. Split into a 2x2 grid (so a pane is neither full width nor full height).
2. In one pane, run commands so that output lines exceed the current viewable area; ie, make use of the scrollback buffer.
3. Use mouse wheel to scroll back to verify that there are scrollback lines.
4. Zoom that pane (cmd-shift-enter)
5. Verify that scrollback is working
6. Un-zoom (cmd-shift-enter again).
7. Notice that it's as if the screen has cleared; basically, looks exactly like what you get from the `clear` command
8. Scroll up - the scrollback content is just empty lines.

### Expected behavior

zoom/un-zoom should not clear the screen and scrollback lines should not be lost.

### Macterm version

Version 1.18.1 (1.18.1)

### macOS version

OS: macOS Tahoe 26.5.1 (25F80) arm64

### Logs or screenshots

https://github.com/user-attachments/assets/e0a62eb8-5af1-4d33-9491-7a5d7f0d617f

BTW, excellent job on this terminal! :)

thdxg:
@armyers Thanks for your feedback! I cannot reproduce this issue on the latest version (v1.19.0). Does the issue persist on your machine with the latest release?

armyers:
Hi @thdxg yes, it still persists on my computer with 1.19.0. Any thoughts on what might be the problem on my system?  The weird thing is that ghostty, cmux and supacode do not have this problem.

thdxg:
Hmm. Does this happen every time you zoom or only sometimes?

armyers:
yes, happens every time. It's not intermittent.

thdxg:
The root cause was u16 underflow in row count calculation, which was fixed in this commit in upstream ghostty: https://github.com/ghostty-org/ghostty/commit/7fa6fffbca802897047186eed7b43faa3bcb87cf

[...]

PR #169, opened 2026-07-14
Terminal-state introspection + control-socket verbs (#165–#167) and a fork-drift guard (#168)

### LINKED ISSUE #165: Add `macterm pane inspect`/`dump` for read-only terminal-state introspection
### Motivation

The control CLI can list panes but can't report **what a pane's terminal core actually contains**. When debugging a scrollback/reflow bug (e.g. #112), the only way to observe the surface's real state is to temporarily add an `os.Logger` to `GhosttyTerminalNSView`, rebuild, and grep `log stream` — a slow loop repeated for every hypothesis. A read-only introspection verb would turn that into a single command.

### Proposal

Add read-only `pane` subcommands that surface libghostty's own state over the existing control socket:

```
macterm pane inspect --pane 1        # grid cols×rows, scrollback (total/offset/len rows),
                                     # cursor row/col, alt-screen flag, content scale,
                                     # foreground pid + argv
macterm pane dump --pane 1                 # visible viewport text
macterm pane dump --pane 1 --scrollback    # full scrollback + viewport text
```

Backed by existing libghostty calls: `ghostty_surface_size()` (grid + cell px), the `GHOSTTY_ACTION_SCROLLBAR` values already wired into `SurfaceScrollView` (`total/offset/len`), `ghostty_surface_foreground_pid`, etc. `--json` like the other verbs.

`inspect` alone would have made #112 a ~2-minute diagnosis (the scrollback `total` jumping 204 → 65706 on un-zoom was the whole signal, and I had to hand-instrument to see it).

### Scope / notes

- Read-only; same same-user socket security model as the rest of the CLI.
- `dump` needs a way to read cell contents out of the core — may need a small libghostty addition if no C API exposes the grid text today (worth checking `ghostty.h`).

### Area

`area:terminal`, CLI (`macterm`).

thdxg:
Four related follow-ups from debugging #112 — the control CLI could not observe or drive a pane's terminal core, so every hypothesis meant hand-instrumenting a build. This adds the introspection/control verbs that turn that into one command, plus a guard against the stale-fork bug that caused #112 in the first place.

## #165 — `pane inspect` / `pane dump`

- **`pane inspect`** — read-only snapshot of libghostty's live state: grid, cell/surface pixels, scrollback `total`/`offset`/`len` (the cached `GHOSTTY_ACTION_SCROLLBAR` values), content scale, foreground pid + argv, and the `processExited`/`needsConfirmQuit` flags. The scrollback totals are exactly the signal #112 needed (a `total` jumping to ~65k on un-zoom was the whole diagnosis).
- **`pane dump [--scrollback]`** — viewport text, or the full scrollback, via `ghostty_surface_read_text` (the same API ghostty's own app uses for select-all).
- **Two honest gaps**, both documented: alt-screen is a heuristic (`total ≤ len`, the signal the scroll code already uses) because the C ABI has no direct query; **cursor position is omitted** because the ABI doesn't expose it — surfaced in docs rather than faked. (A `ghostty_surface_cursor` addition to the fork could fill both later.)

## #166 — `pane zoom` / `pane resize-split`

- **`pane zoom`** toggles zoom via the same `TerminalTab.toggleZoom` the keybind uses — deterministic, no `osascript`.
- **`pane resize-split --axis h|v --ratio R`** sets the nearest matching-axis branch ratio absolutely (new `SplitNode.settingRatio`, mirroring the existing `applyResize` tree walk), for reproducing an exact geometry.

## #167 — debug-only `pane resize`

- **`pane resize --cols C --rows R`** drives a single in-place `ghostty_surface_set_size`, bypassing SwiftUI layout, to isolate a resize/reflow transition. **Gated `#if DEBUG` in both the CLI and `ControlHandler`'s dispatch** — a release app answers `unknown_command`, so it never ships live. (The live layout reasserts the real geometry on the next tick, so it drives the *transition* — and any reflow side effect, visible in `inspect`'s scrollback — rather than persisting a size; documented.)

## #168 — fork-drift guard

- **This repo:** `scripts/setup.sh` now reads libghostty's baked `build_config.version`, resolves the fork commit's parent (the pristine upstream base), and **warns** when that base is materially behind `ghostty-org/ghostty` HEAD. Strictly advisory — any failure to determine it (offline, no `gh`, unparseable) is skipped so setup always completes.
- **Forks (separate PRs, I have admin):** a scheduled `assert-no-drift.yml` that asserts the invariant the rewritten `sync-upstream.yml` maintains — `main` = pristine upstream + exactly one downstream commit touching only the declared allowlist, within a freshness window — failing loudly on any other divergence:
  - thdxg/ghostty#4
  - thdxg/zmx#2
  - Validated positive + negative against both live forks under real bash. `workflow_dispatch` can't run pre-merge (GitHub requires the workflow on the default branch first).

## Wire protocol

Additive only: new `ControlArgs` fields (`scrollback`, `axis`, `ratio`; reuses `rows`/`cols`) and two `ControlData` structs (`ControlPaneInspect`, `ControlPaneDump`). No new error codes — reuses `no_surface`/`bad_request`/`not_found`. Old clients ignore unknown fields, as before.

## Testing

- `mise run format` / `lint` / `test` all green (Release build too, confirming the `#if DEBUG` gating doesn't break release).
- New handler + codec tests cover selector resolution, arg validation, the zoom/ratio model, and the debug-verb gate.
- Headless panes have no surface, so the live-surface paths (`inspect` dims, `dump`, debug `resize`) were **verified end-to-end against a running debug app**: `inspect` reported the live 132×65 grid and scrollback flipping 65→360 as output scrolled; `dump`/`--scrollback` round-tripped viewport vs full history; `zoom`, `resize-split`, and the debug `resize` all drove their effects and validation errors as expected.

Closes #165, #166, #167. Addresses #168 (setup.sh guard here; fork-side guards in the linked fork PRs).

## Summary by CodeRabbit

* **New Features**
  * Added pane inspection and terminal text dump commands.
  * Added commands to toggle pane zoom and adjust split ratios.
  * Added read-only pane diagnostics, including scrollback and process details.
  * Added debug-only pane resizing for troubleshooting layout transitions.

* **Documentation**
  * Updated CLI documentation with the new commands, outputs, and debug-only behavior.
  * Updated contributor guidance to reference the current CLI documentation.


thdxg:
@coderabbitai review plz

PR #170, opened 2026-07-14
Show the running AI agent's logo as the sidebar tab icon

OpenSource03:
## What

A tab whose pane is running an AI coding agent CLI — Claude Code, Codex, opencode, Cursor, Gemini CLI, Copilot CLI, or Grok — now shows that agent's logo in the sidebar instead of the generic tab icon, tinted with the vendor's brand color (Anthropic coral, OpenAI light gray, Google blue, GitHub purple; monochrome brands stay `.primary`). A new **Show AI agent icons** toggle in Settings → Appearance → Sidebar turns it off (on by default).

CleanShot 2026-07-14 at 19 05 53

## Why

With several agent sessions open at once, every tab looks the same. The pane already knows its foreground process — surfacing *which* agent runs where makes agent-heavy workflows scannable at a glance, the same way the process-derived tab title does for names.

## How

- `AgentIcon` (`Macterm/Model/AgentIcon.swift`) maps a pane's foreground process to a bundled logo. Matching can't rely on the kernel `comm` alone: the binary's name often isn't the CLI name (brew's codex is `codex-aarch64-apple-darwin`, comm-truncated to `codex-aarch64-a`; Claude Code's native install execs a versioned binary whose comm is the bare version, e.g. `2.1.207`). So matching is exact, then boundary-aware prefix on comm (`codex-aarch64-a` → codex, but `claudette` ≠ claude), then an argv[0]-basename fallback read lazily via the same resolved-pid path the title pipeline uses (zmx-aware after the merge with main).
- The result is cached on `Pane` per foreground pid/comm inside `applyForegroundRefresh`, so the existing 250ms poll republishes only on change and argv is read only when the foreground actually changes.
- Logos are monochrome template SVGs (LobeHub icon set) in a new `Assets.xcassets`, rendered like SF Symbols; `SidebarRowIcon` prefers a live agent logo over the user's chosen tab icon and composes with the status indicator (running spinner still replaces it, the done dot badges it).
- The toggle follows the settings pattern: `Preferences.showAgentIcons` (default true) gating only the sidebar row, not the poll.

## Verified

- [x] Built and ran the change in the app (`mise run run`) and confirmed the logos appear/expire live for Claude Code (native install), Codex (brew), and fall back to the normal icon for shells and ordinary programs
- [x] Added `AgentIconTests` covering name matching (exact, suffixed, separator-gated prefix), the argv[0] fallback, per-pid caching, and tab-level focused-pane preference

## Notes for reviewers

- Brand tints are hardcoded vendor identity colors — a deliberate exception to "all colors come from MactermTheme", called out in a comment.
- Agents launched as a bare `node`/`bun` script (some npm installs) aren't detected, since both comm and argv[0] are the interpreter; that's a possible follow-up (peek at argv[1] when argv[0] is an interpreter).
- Remote panes report no local foreground pid, so they simply never show an agent logo.

## Summary by CodeRabbit

- **New Features**
  - Added AI agent logo support to terminal sidebar tabs, with an optional toggle in Appearance to show/hide the logos.
  - Added branded logos for additional agents, including Pi.
- **Bug Fixes**
  - Improved agent detection to more reliably identify the running agent across process naming and interpreter-backed commands, reducing false matches.
- **Tests**
  - Added automated coverage for agent-icon matching logic, foreground caching behavior, tab selection, and invoked-name handling.


thdxg (commented):
@OpenSource03 Thanks for this! This is a really neat feature. A few addressable items:

**1. argv fallback runs on the default poll for every non-agent process.** In `AgentIcon.match(comm:argv0:)`, a non-matching `comm` always invokes the `argv0` closure → a `KERN_PROCARGS2` read. That fires on the first foreground frame of *any* non-agent process (bash, fish, node, hx…), including when the status indicator is off — the path `AppState.refreshAllForegroundProcesses` documents as deliberately argv-free. Gate the fallback to only the case it exists for (opaque/versioned `comm`):

```swift
static func match(comm: String?, argv0 fallback: () -> String?) -> AgentIcon? {
    if let icon = match(processName: comm) { return icon }
    // argv[0] only rescues an opaque comm (e.g. a versioned binary); a real
    // process name that simply isn't an agent shouldn't cost a syscall.
    guard comm == nil || ProcessInspector.looksLikeVersionString(comm!) else { return nil }
    return match(processName: fallback())
}
```

(If you'd rather keep `AgentIcon` free of a `ProcessInspector` dependency, pass a precomputed `commIsOpaque: Bool` from `applyForegroundRefresh` instead.)

**2. Redundant pid resolution.** `foregroundArgv0Basename(forPane:)` re-resolves the foreground pid, but `refreshForegroundProcess` already resolved it one line above and passes it as `foregroundPID:`. When the fallback fires, a wrapped pane resolves the pid twice for the same frame (and the two could disagree across the gap). Fixing #1 hides most of this; cleaner is to capture the already-resolved pid in the closure. Minor.

**3. Add a `CaseIterable` guard test.** `processNames` is hand-maintained and the enum isn't `CaseIterable`, so adding an agent case + asset but forgetting the dict entry is a silent no-op. Making `AgentIcon: CaseIterable` + a test asserting every case has a process-name mapping is cheap insurance.

**4. Confirm the new `Assets.xcassets` is in the target.** This is the app's first asset catalog; if `project.yml`'s sources don't pick it up, `Image("AgentClaude")` renders an empty view with no crash or log, and the logic-only unit tests won't catch it. Worth a clean `xcodegen` + visual check per agent.

Non-blocking: the `15×15` frame is fixed while the sibling SF Symbols scale with font/Dynamic Type — consider `@ScaledMetric` so the logo scales with the row.

thdxg:
@OpenSource03 Could you also add icon for Pi agent?

OpenSource03:
@thdxg Thanks for the review! All addressed in a878aba:

1. Gated the fallback like you suggested — only runs for version-shaped or interpreter comms now, with a test asserting it never fires for normal process names.
2. Fixed — pid is resolved once and reused in the closure.
3. Done, with the guard test.
4. Checked — all icons land in the compiled catalog and I've verified them visually in the app.
5. Done.

Pi added too.

thdxg:
@OpenSource03 Thanks for your work! LGTM

PR #171, opened 2026-07-14
Multi-select sidebar rows to remove projects and tabs in bulk

OpenSource03:
## What

The sidebar now supports multi-selection: Cmd/Shift-click a bunch of projects and/or tabs, right-click, and close or remove them all in one go.

## Why

Cleaning up the sidebar meant right-click → close, one row at a time. Once you've got a few projects with a dozen tabs each, that gets old fast. Bulk delete is the obvious missing affordance.

## How

Two changes, really:

- `selection` becomes a `Set<SidebarItem>` instead of an optional. That alone gives `List` native Cmd/Shift-click multi-select. The navigation `onChange` now only follows single selections, so extending a selection doesn't keep yanking the active tab around.
- The per-row `.contextMenu`s move into a single list-level `.contextMenu(forSelectionType:)`. A per-row menu can't see the selection, so this is the only way to make "right-click the selection → act on all of it" work. SwiftUI hands the closure exactly the set to act on — the whole selection when the click lands inside it, just the clicked row otherwise. The single-row menus (New Tab, Copy Path, Rename, Move to Project, Unload…) are reproduced as-is, and a multi-selection gets one destructive item: "Close N Tabs", "Remove N Projects", or "Remove N Items" for a mix.

Destructive paths keep the busy-pane confirmations: single-row Close/Remove/Unload still go through `requestCloseTab` / `requestRemoveProject` / `requestUnloadProject`, so a plain right-click behaves exactly as before. Bulk delete gets its own `requestRemoveSelection` (same pending/alert pattern, alert wired up in `MactermApp`) so a selection with something running prompts once for the whole batch instead of once per item. Tabs whose project is also selected are skipped — `removeProject` already tears them (and their zmx sessions) down.

`AppState.removeProjects` / `closeTabs` are just thin loops over the existing single-item methods.

## Verified

- [x] Built and ran the change in the app (`mise run run`): Cmd/Shift-click builds a selection, right-clicking inside it shows the bulk action, right-clicking an unselected row shows the normal single-row menu
- [x] Added tests for `removeProjects`, `closeTabs`, and `requestRemoveSelection` running its removal inline when nothing is busy

## Summary by CodeRabbit

* **New Features**
  * Added multi-selection support in the sidebar for removing multiple projects or tabs at once.
  * Added confirmation prompts before removing items with running processes.
  * Added options to confirm or cancel pending bulk removals.

* **Bug Fixes**
  * Bulk removals and tab closures now save changes consistently after the entire operation.
  * Prevented tabs from being closed redundantly when their projects are removed.


thdxg:
@OpenSource03 Thanks for your work! LGTM

PR #172, opened 2026-07-14
Fix panes failing to spawn from trailing-slash project paths (fatal under nushell; blank prompt dir under zsh)

OpenSource03:
## What

New panes in a local project spawn with a blank directory in the zsh prompt until the first `cd`. Cause: project paths are stored with a trailing slash. This PR strips it.

**Before → after** (first prompt in a fresh pane):

```
➜                          ➜  my-project
➜  git:(main)              ➜  my-project git:(main)
```

## Why it happens

```
Open panel  ──▶  URL.path(percentEncoded:)     keeps the slash: "/Users/me/dev/my-project/"
            ──▶  projects.json                 stores it verbatim
            ──▶  libghostty working_directory  exports it as $PWD
            ──▶  zsh adopts inherited $PWD     "/Users/me/dev/my-project/"
            ──▶  prompt %c / %1~               last component of ".../my-project/" = ""  ← blank
```

One-line repro:

```
$ env PWD="/Users/me/dev/my-project/" zsh -f -c 'print -rP "pc=[%c]"'
pc=[]
```

The first `cd` makes zsh recompute `$PWD` normalized — which is why the prompt "fixes itself".

## Fix

New `ProjectPath.normalizedForStorage(_:)` — locals through the existing `canonicalLocal` (slash stripped, `~` expanded, `.`/`..` standardized); remote specs pass through verbatim.

| Where | Covers |
|---|---|
| `ProjectStore.add` / `setPath` | all new paths: open panel, remote sheet, CLI `project create`, Replace Project Path |
| `ProjectStore.load()` | existing `projects.json` — migrates on next launch, persists clean on next save |
| `GhosttyTerminalNSView` spawn (local branch only) | panes restored from old workspace snapshots that persisted a slashed path |

## Verified

- [x] `format`, `lint`, full test suite (680 passing)
- [x] Ran the app: pane in a previously slash-stored project now starts with a clean `$PWD`, prompt shows the directory immediately
- [x] New tests: normalization helper, `add`/`setPath`, remote verbatim, `load` migration against a hand-crafted pre-fix `projects.json`

## Notes

- Already-running zmx shells keep their old `$PWD` until the user `cd`s once — reattach hands back a live process; nothing safe to do there.
- Old workspace snapshots aren't rewritten; the spawn-time canonicalization covers them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Project paths are now stored in a consistent canonical format, including trailing-slash removal and `~/` expansion for local paths.
  * Existing saved projects are automatically normalized after loading (in memory) to match the current storage format.
  * Remote-style paths are preserved exactly, including any trailing slash.
  * Restored terminal sessions now use the correct working directory derived from the normalized stored path.

* **Tests**
  * Expanded test coverage for path normalization across adding, updating, loading legacy data, and handling local vs. remote paths.

---

## Update (maintainer): this bug is fatal under nushell, not just cosmetic

The trailing slash does more than blank a zsh prompt segment. libghostty exports `working_directory` **verbatim as $env.PWD**, and nushell refuses to start on a non-canonical value:

```
Error: nu::shell::error
  × $env.PWD contains trailing slashes
  crates/nu-protocol/src/engine/engine_state.rs
```

With nu as the login shell, every project added via **Cmd+O fails to spawn its pane** — the whole `zmx → login → bash → exec -l nu` chain collapses in ~150 ms and ghostty shows its abnormal-exit screen (issue reported as \"Ghostty failed to launch the requested command … Runtime: 145 ms\"). The palette path builds paths with `appendingPathComponent` (no slash), which is why only the open panel was affected.

Verified end-to-end in a hermetic app instance (throwaway $HOME + `MACTERM_BENCHMARK_DATA_DIR`, driven over the control socket): a projects.json seeded with a trailing-slash path reproduces the exact error screen pre-fix and spawns a live nu pane post-fix.

cd748ec strengthens the fix:
- Spawn-time guard uses `normalizedForStorage` instead of bare `canonicalLocal`, so a non-local-shaped string passes through untouched instead of being coerced relative to the app's own cwd
- Comments document the fatal mechanism at both layers
- New test pins every non-canonical form (`//`, `/./`, `/../`, `~/`) — nu validates the whole string, not just trailing slashes

### DIFF
--- a/Macterm/Model/ProjectPath.swift
+++ b/Macterm/Model/ProjectPath.swift
@@ -98,6 +98,18 @@ enum ProjectPath: Equatable {
         return standardized
     }
 
+    /// Storage form of a raw project path: local paths canonicalized (tilde
+    /// expanded, `.`/`..` standardized, trailing slash stripped); remote specs
+    /// and unparseable strings pass through verbatim. A stored trailing slash
+    /// would otherwise reach the spawned shell's `$PWD` verbatim — zsh adopts
+    /// the inherited value when it names the cwd, and `%c`/`%1~` render the
+    /// empty last component of `/path/dir/`, blanking the prompt's directory
+    /// segment until the first `cd`.
+    static func normalizedForStorage(_ raw: String) -> String {
+        if case let .local(path)? = parse(raw) { return canonicalLocal(path) }
+        return raw
+    }
+
     /// Convenience for call sites holding a raw path string (`Project.path`,
     /// `Pane.projectPath` — the remote spec travels and persists as a string).
     static func isRemote(_ raw: String) -> Bool {
--- a/Macterm/Model/ProjectPath.swift
+++ b/Macterm/Model/ProjectPath.swift
@@ -101,10 +101,13 @@ enum ProjectPath: Equatable {
     /// Storage form of a raw project path: local paths canonicalized (tilde
     /// expanded, `.`/`..` standardized, trailing slash stripped); remote specs
     /// and unparseable strings pass through verbatim. A stored trailing slash
-    /// would otherwise reach the spawned shell's `$PWD` verbatim — zsh adopts
-    /// the inherited value when it names the cwd, and `%c`/`%1~` render the
-    /// empty last component of `/path/dir/`, blanking the prompt's directory
-    /// segment until the first `cd`.
+    /// would otherwise reach the spawned shell's `$PWD` verbatim (libghostty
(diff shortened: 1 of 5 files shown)


thdxg:
@OpenSource03 Thanks for your work! I actually discovered that the trailing slash makes nushell reject the path and fail to start while reviewing this. I added a small modification to this PR. LGTM!

PR #175, opened 2026-07-20
Add Shift+Enter to search for downward navigation; fix inverted chevrons

jcuhnoio:
## Summary

- **Shift+Enter** in the Cmd+F search bar now navigates down the scrollback; plain **Enter** keeps navigating up.
- Fixes the chevron buttons, which were wired backwards: the up-chevron scrolled down and vice versa.

## Direction semantics

Verified against ghostty source (upstream and the `thdxg/ghostty` fork): `navigate_search:next` walks matches newest→oldest — i.e. **up** the scrollback — and `previous` walks back down. Ghostty's own macOS app maps the standard Find Next to `next` the same way.

So Enter's existing `onSubmit { onNavigateNext() }` already scrolled up; this PR adds a `.onKeyPress(keys: [.return])` handler (same pattern as the bar's Escape/Cmd+F handlers, which already fire while the field is focused) that sends `onNavigatePrevious()` when shift is held, and swaps the chevron buttons to agree with the actual scroll direction.

## Testing

- `mise run format`, `lint`, and the full test suite pass. (Search bar is SwiftUI view code, which is intentionally not unit-tested per repo conventions.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the search navigation chevron buttons so they move to the expected next and previous results.
  * Added Shift+Return support to navigate to the previous search result while leaving Return navigation unchanged.

### DIFF
--- a/Macterm/Views/SearchBar.swift
+++ b/Macterm/Views/SearchBar.swift
@@ -37,10 +37,10 @@ struct TerminalSearchBar: View {
                 .clipShape(RoundedRectangle(cornerRadius: 6))
                 .overlay(RoundedRectangle(cornerRadius: 6).strokeBorder(MactermTheme.border, lineWidth: 1))
 
-                Button(action: onNavigatePrevious) {
+                Button(action: onNavigateNext) {
                     Image(systemName: "chevron.up").font(.system(size: 10, weight: .semibold))
                 }.buttonStyle(SearchButtonStyle())
-                Button(action: onNavigateNext) {
+                Button(action: onNavigatePrevious) {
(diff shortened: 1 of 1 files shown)


jcuhnoio:
@thdxg 
I'm not entirely sure if the performance benchmarks are a reliable metric - just off my intuition, though. I don't think the performance differences make sense given that it's a small change. (I could be wrong)

thdxg:
@jcuhnoio Yeah the benchmark has a lot of noise, which is why I only made it flag when the delta is above threshold. This pr wasn't flagged, so no impact on performance

thdxg:
Lgtm

PR #176, opened 2026-07-20
Show search matches as scrollbar tick marks, colored like the text highlights

jcuhnoio:
## Summary

Cmd+F now marks every match's position in the scrollbar track (Xcode-style ticks), so you can see where matches live in the scrollback at a glance. The selected match's tick is drawn last and in the selected-highlight color, so it stands out from candidate matches.

## Design

**libghostty never reports match positions** — its search API only emits counts (`SEARCH_TOTAL` / `SEARCH_SELECTED`). So tick rows are recovered Macterm-side:

- **`SearchTicks`** (pure, unit-tested): reads the scrollback via the existing `ghostty_surface_read_text` path and scans it with the same ASCII-case-insensitive literal match ghostty's search uses (`std.ascii.indexOfIgnoreCase` in `sliding_window.zig`), re-wrapping logical lines at the surface's column count to map matches to visual rows. The match list stays 1:1 with ghostty's, so the `SEARCH_SELECTED` index (counted from the newest match) picks the selected tick; if the counts ever disagree, the selected highlight is dropped rather than drawn on the wrong tick.
- **`SearchTickOverlay`**: an `NSView` pinned to the scroller track in `tile()`, above the overlay scroller so ticks stay visible while it fades out, with hit-testing disabled so scrollbar click/drag is untouched.
- **Wiring**: `SurfaceScrollView` re-scans on `SEARCH_TOTAL` updates (coalesced 100ms — totals stream while ghostty walks history), recolors the selected tick on `SEARCH_SELECTED`, clears on search end. The scan runs against the needle actually sent to the core, not the live (possibly newer) field text.

**Tick colors mirror the renderer's search highlight backgrounds** (`search-background` / `search-selected-background`, defaults golden yellow `#FFE082` / soft peach `#F2A57E`), so ticks read as the same colors as the highlighted text. Those keys are `TerminalColor` unions with no `cval`, so `ghostty_config_get` can't return them — a new pure `SearchHighlightColors` helper scans the user's config text (same pattern and config-file-include limitation as `ShellIntegrationFeatures`), honoring hex overrides and falling back to the defaults for named-X11 / `cell-*` values a tick can't reproduce.

## Known approximations (noted in code comments)

- Columns are counted as UTF-8 code points, so double-width cells can drift a tick by one row — invisible at scrollbar scale.
- A match spanning a hard line break isn't found (ghostty itself is ambiguous there).

## Testing

- Unit tests for `SearchTicks` (wrap arithmetic, case-insensitivity, non-overlap, selected-from-end mapping, multibyte counting) and `SearchHighlightColors` (defaults, hex/quoted parsing, last-wins, fallback semantics).
- `mise run format`, `lint`, and the full suite pass. Overlay drawing is AppKit view code, intentionally not unit-tested per repo conventions.

## Summary by CodeRabbit

* **New Features**
  * Added scrollbar markers showing the locations of terminal search matches.
  * Highlighted the currently selected match with a distinct marker.
  * Search markers update as queries, selections, and scrollback content change.
  * Marker colors now follow configured search highlight colors, with sensible defaults.

* **Bug Fixes**
  * Improved search matching across wrapped lines, mixed case, and multibyte text.

### DIFF
--- a/Macterm/Config/SearchHighlightColors.swift
+++ b/Macterm/Config/SearchHighlightColors.swift
@@ -1,0 +1,57 @@
+import Foundation
+
+/// Resolves the colors ghostty's renderer paints search highlights with
+/// (`search-background` for candidate matches, `search-selected-background`
+/// for the focused one), so the scrollbar ticks match the highlighted text.
+///
+/// These keys are `TerminalColor` unions in ghostty and that union has no
+/// `cval`, so `ghostty_config_get` returns false for them — the resolved
+/// values can't be read through the C API. Instead the user's config text is
+/// scanned directly, with the same pattern and limitation as
+/// `ShellIntegrationFeatures`: last occurrence wins, and a value set in a
+/// `config-file` include isn't seen.
+enum SearchHighlightColors {
+    struct RGB: Equatable {
+        let r, g, b: UInt8
+    }
+
+    /// Ghostty's documented defaults (`src/config/Config.zig`): candidate
+    /// matches on golden yellow, the selected match on soft peach.
+    static let defaultMatch = RGB(r: 0xFF, g: 0xE0, b: 0x82)
+    static let defaultSelected = RGB(r: 0xF2, g: 0xA5, b: 0x7E)
+
+    static func matchBackground(inConfigText text: String?) -> RGB {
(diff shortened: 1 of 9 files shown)


jcuhnoio:
Here's an example of what it looks like:
image

PR #178, opened 2026-07-21
Add adaptive and configurable terminal backgrounds

OpenSource03:
## What

Adds an opt-in Appearance setting that matches full-screen terminal apps such as OpenCode: a single pane can tint the whole window, while split panes adapt independently and leave the surrounding chrome unchanged.

## Why

Full-screen TUIs can paint a background that clashes with transparent terminal padding, the sidebar, titlebar, and Liquid Glass.

### Properly adapts to split view
CleanShot 2026-07-21 at 02 09 54
CleanShot 2026-07-21 at 02 10 11

### With only single pane, whole app adapts to the TUI color
CleanShot 2026-07-21 at 02 10 39

### With feature off
CleanShot 2026-07-21 at 02 11 39

## How

- Keeps the Ghostty config as the source of truth for terminal colors; the adaptive tint is a transient presentation layer on top of it.
- Prefers explicit OSC 11 background changes, with bounded IOSurface sampling and two-observation stabilization for apps that do not report a color.
- Samples every visible pane independently. Single-pane tabs publish a window-wide tint; split tabs fill only matching panes.
- Routes chrome colors through MactermTheme and selects semantic light/dark appearance using WCAG contrast.
- Performs no sampling and owns no monitoring timer while disabled.

## Verified

- [x] Repository format, lint, and full test scripts pass after merging current upstream main
- [x] Built and ran the change in the app
- [x] Confirmed Grok adapts without switching tabs
- [x] Confirmed split-pane colors remain pane-local while alternating focus
- [x] Confirmed disabling the setting restores the configured theme
- [x] Added focused tests for contrast, IOSurface detection, stabilization, and split presentation

## Notes for reviewers

The feature is off by default. IOSurface is linked explicitly because sampling reads the renderer-owned BGRA8 surface directly; it does not use screen capture or request additional permissions.

## Summary

* **New Features**
  * Added adaptive terminal chrome to mirror app background tints, including temporary per-pane adaptive background fills.
  * Added terminal background settings to choose background source (Ghostty-resolved vs custom) and edit a remembered custom override color.

* **Bug Fixes**
  * Improved adaptive/background update synchronization with debounced config reloads.
  * Refined unfocused split dimming to keep adaptive-background panes color-accurate.
  * Updated background/color change handling to refresh UI promptly from Ghostty updates.

* **Tests**
  * Added/expanded coverage for adaptive detection, override parsing/generation, and theme/contrast math.


Macterm/Model/SplitNode.swift
```diff
@@ -339,6 +339,10 @@ final class Pane: Identifiable {
     private var agentIconPID: pid_t?
 
     let searchState = TerminalSearchState()
+    /// Temporary full-pane fill used when a TUI owns this leaf's background in
+    /// a split. It is presentation-only, never persisted, and intentionally a
+    /// CGColor so the model remains independent of AppKit and SwiftUI.
+    var adaptiveBackgroundColor: CGColor?
```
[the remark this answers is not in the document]
OpenSource03: Fixed in f53f6f6. The terminal leaf now reports color changes through an explicit callback: workspace panes route through AppState, while quick-terminal panes route through their separate QuickTerminalSplitState owner. Added an AppState ownership regression test.

MactermTests/Ghostty/AdaptiveTerminalBackgroundTests.swift
```diff
@@ -0,0 +1,157 @@
+import AppKit
+@preconcurrency import IOSurface
+@testable import Macterm
+import Testing
+
+struct AdaptiveTerminalBackgroundTests {
```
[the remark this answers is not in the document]
OpenSource03: Fixed in f53f6f6 by annotating AdaptiveTerminalBackgroundTests with @MainActor.

thdxg:
@OpenSource03 Wow this is impressive. Thanks for your work! I applied some fixes for issues I encountered while testing this locally:

1. Theme-background flash when switching to a TUI tab. pruneState wiped a pane's detected color when its tab went off-screen, so revisiting restarted detection (~0.25–0.75s of configured background). Now remembered colors survive occlusion and re-seed the stabilizer, so a revisited tab presents instantly; a TUI that exited off-screen still clears via the normal two-observation path.

2. Translucent padding seam around a lone TUI pane. Single-pane tabs only got the window-wide tint, which follows windowOpacity — leaving a see-through gap next to the TUI's opaque pixels. Every detected pane now gets the same opaque fill splits already had; the window tint rule is unchanged.

3. Regression: Settings window reopening on activation + broken Cmd+D/Cmd+W. The adaptive tint fed .preferredColorScheme, which is scene-level (it also governs the Settings window). It flapped on every tint adopt/clear, destabilizing SwiftUI window management and breaking the cached window identity that gates hotkeys. The app-wide scheme now derives from the config background only (keeping the WCAG classification); window-scoped tint and text contrast are untouched.

MactermTests/Ghostty/ThemeTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// adopted or cleared, which destabilized SwiftUI window management (the
+    /// closed Settings window reopened on every app activation, and hotkeys
+    /// broke because the WindowGroup window lost its cached identity).
+    @Test
+    func colorSchemeIgnoresTheTransientAdaptiveTint() {
+        let configScheme = MactermTheme.colorScheme
+        let opposite: NSColor = configScheme == .dark ? .white : .black
+
+        GhosttyApp.shared.adoptAdaptiveBackgroundColor(opposite)
+        defer { GhosttyApp.shared.adoptAdaptiveBackgroundColor(nil) }
```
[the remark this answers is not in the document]
thdxg: Fixed in 0e16f2e. The test now captures the prior adaptive tint and restores it in the defer instead of clearing to nil.

thdxg:
@OpenSource03 The improvements are applied but I want to discuss whether we need the background override setting. Ghostty already has a background setting in its config, and I want to avoid having duplicate responsibilities between ghostty and macterm as much as possible. Is there a reason you wanted the override setting?

OpenSource03:
> @OpenSource03 The improvements are applied but I want to discuss whether we need the background override setting. Ghostty already has a background setting in its config, and I want to avoid having duplicate responsibilities between ghostty and macterm as much as possible. Is there a reason you wanted the override setting?

I could see that from the way it was handled originally, yet I would, personally, much more prefer to have a color picker UI than having to edit Ghostty config. I did deliberately leave option to use Ghostty config, but I find it much easier to edit stuff via UI than via some config file, specially when it comes to colors.

Additionally, if we go with such logic, isn't the background blur something Ghostty controls too?

thdxg:
@OpenSource03 Yes Macterm does have its own opacity and blur settings which could seem redundant with ghostty config. But those are slightly different from ghostty's other appearance settings, because opacity and blur of the terminal surface needed to be always overridden to zero, to completely show through the app window's opacity and blur. Otherwise, the terminal panes' appearance can diverge from rest of the app's.

There are a few more settings ignored from ghostty config (documented [here](https://macterm.thdxg.dev/docs/configuration)) because Macterm explicitly overrides. Since ghostty config doesn't control these, I added them in the settings UI.

When it comes to background color or any other color customization (e.g. foreground, selection-background, selection-foreground, etc.), Macterm still respects ghostty config for them. So unlike window opacity/blur, creating UI to edit those settings introduces two distinct ways to achieve the same thing.

thdxg:
@OpenSource03 I'm happy to discuss more about the settings. But either way, I think it would have to be a separate discussion from this PR, as this mainly implements appearance adapting to fullscreen TUI. Would you be okay with removing the settings change in this PR and continue discussing in another?

OpenSource03:
> @OpenSource03 I'm happy to discuss more about the settings. But either way, I think it would have to be a separate discussion from this PR, as this mainly implements appearance adapting to fullscreen TUI. Would you be okay with removing the settings change in this PR and continue discussing in another?

Hey @thdxg, sorry for the late reply. Yes, I can see your point, and it does make sense. I do also see benefit of having the override option with a color picker, but I will leave it up to you to decide. 

I will remove the color picker from this PR probably later today or tomorrow.

OpenSource03:
@thdxg Done! :)

Macterm/Ghostty/GhosttyApp.swift
```diff
@@ -510,6 +518,18 @@ final class GhosttyApp {
...
 
+    /// Update the temporary chrome tint without touching libghostty's config.
+    /// The terminal renderer remains authoritative for its own pixels; this
+    /// only brings the native window surfaces into visual alignment.
+    func adoptAdaptiveBackgroundColor(_ color: NSColor?) {
+        if let current = adaptiveBackgroundColor, let color, current.isVisuallyEqual(to: color) {
+            return
+        }
+        guard adaptiveBackgroundColor != nil || color != nil else { return }
+        adaptiveBackgroundColor = color
```
thdxg, line 607: **Blocking: SwiftUI chrome has no observation path to the tint.**

Compare `adoptResolvedColors` a few lines above — it does `configVersion += 1` *and* posts. This only posts.

The notification reaches AppKit observers (`WindowAppearance.sync`, quick-terminal blur), but SwiftUI chrome reads `MactermTheme.bg`/`fg`/`bgWithOpacity`, which are plain `static var` computed accessors carrying no observation of their own. `AppColorScheme` is the only thing establishing that dependency, via `_ = GhosttyApp.shared.configVersion`.

`adaptiveBackgroundColor` *is* observable (not `@ObservationIgnored`), but nothing in the SwiftUI tree reads it — and MactermApp.swift now carries a comment saying it deliberately does *not* track it. So the sidebar and palette keep the old theme color until some unrelated event bumps `configVersion`. The screenshots showing the sidebar adapting suggest it currently works only because render activity keeps firing other config changes, i.e. accidentally.

Either bump `configVersion` here, or have something in the tree read `adaptiveBackgroundColor`. Worth saying which you intend, since the existing comment argues against the latter.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
```diff
@@ -0,0 +1,249 @@
...
+        var stabilizer = stabilizers[id]
+            ?? AdaptiveTerminalBackgroundStabilizer(seededWith: view.sampledDominantBackgroundColor)
+
+        if let reported = effectiveCandidate(view.reportedBackgroundColor) {
+            stabilizer.reset(to: reported)
+            stabilizers[id] = stabilizer
+            return false
+        }
+
+        let candidate: NSColor? = if let surface = view.layer?.contents as? IOSurface {
```
thdxg, line 157: **Blocking: this reads a libghostty implementation detail, not a contract.**

Macterm hands the raw NSView pointer to `ghostty_surface_new` (GhosttyTerminalNSView.swift:230) and libghostty creates and owns the layer. This reaches into that layer's `contents` and assumes an `IOSurface` in BGRA8. Citing "Ghostty's Metal Target.zig" is honest, but it's referencing upstream Zig internals from Swift across a C ABI that makes no such promise.

Given this repo's history — the fork-sync breakage and the #112 reflow overflow — an upstream change to the drawable path or pixel format makes this silently stop working, or worse, misinterpret bytes. The test suite won't catch it either, since it builds its own synthetic `IOSurfaceCreate`.

Minimum fix: guard the pixel format with `IOSurfaceGetPixelFormat(surface)` against BGRA rather than relying only on `bytesPerElement >= 4`, and record the fragility in CLAUDE.md's Known Limitations.

Better: drop pixel sampling for now and ship the OSC 11 path alone — see the summary comment.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
```diff
@@ -0,0 +1,249 @@
...
+            MainActor.assumeIsolated {
+                self?.sampleTimer = nil
+                self?.sampleVisiblePanes()
+            }
+        }
+        RunLoop.main.add(timer, forMode: .common)
+        sampleTimer = timer
+    }
+
+    private func startMonitoring() {
```
thdxg: **Blocking: this timer never stops while the feature is enabled.**

`startMonitoring()` guards on `monitorTimer == nil` and is called from `preferenceDidEnable`, `focusDidChange`, and `terminalDidRender` — but nothing tears it down except `preferenceDidDisable`. So once enabled, the app samples every visible pane twice per second forever, including a fully idle app with no TUI running. Each sample locks the IOSurface and reads ~4000 pixels per pane, contending with the renderer on its own live surface.

This contradicts the design this codebase already settled on for exactly this problem: `PollCadence` bursts, backs off, and stops entirely when nothing is on screen, specifically to avoid polling a static app.

It's also likely to move the `workload-focused`/`workload-unfocused` CPU and wakeups/s metrics in the benchmark harness that just landed (#183) — but since the feature is off by default the benchmark won't flag it, so the regression ships invisibly for anyone who enables it.

Suggest mirroring `PollCadence`: stop after N consecutive no-change samples, resume on a render or focus event.

Macterm/Views/WindowAppearance.swift
```diff
@@ -253,7 +253,7 @@ enum WindowAppearance {
         guard glassSupported else { return }
         if #available(macOS 26.0, *) {
             guard let glass = existingGlass(in: window) else { return }
-            glass.updateKeyStatus(window.isKeyWindow, backgroundColor: GhosttyApp.shared.backgroundColor)
+            glass.updateKeyStatus(window.isKeyWindow, backgroundColor: MactermTheme.nsBg)
```
thdxg: **Blocking (inconsistency): the glass tint helper still uses the heuristic this PR replaces.**

This call now correctly reads `MactermTheme.nsBg`, but the helper it feeds (WindowAppearance.swift:175-176) computes `isLightColor` and `adjustingSaturation` from a plain `luminance > 0.5` threshold — the exact heuristic this PR replaces elsewhere with WCAG contrast via `ThemeColorMath.prefersDarkForeground`.

So with a vivid TUI background, the glass overlay opacity is chosen by the misclassifying branch while `nsFg` uses the correct one. That's visible on precisely the saturated red/magenta cases `semanticForegroundUsesContrastRatherThanRawBrightness` was added to cover. Route `isLightColor` through the new contrast math too.

Separately: `syncKeyStatus` fires on `windowDidBecomeKey`/`windowDidResignKey`, which `.mactermConfigDidChange` does not trigger — worth confirming the tint is picked up on a plain tint change with no focus event, given the `configVersion` gap noted on GhosttyApp.swift.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
```diff
@@ -0,0 +1,249 @@
...
+        guard let window = view.window,
+              window.occlusionState.contains(.visible),
+              !view.isHiddenOrHasHiddenAncestor,
+              view.bounds.width > 0,
+              view.bounds.height > 0
+        else { return false }
+        return true
+    }
+
+    private func effectiveCandidate(_ color: NSColor?) -> NSColor? {
```
thdxg, line 246: **Wrong baseline on split themes.**

The 0.04 threshold suppresses colors too close to the configured background — but `GhosttyApp.shared.backgroundColor` is app-global and, per its own doc comment, "always collapse[s] a split to its light side" for `theme = light:X,dark:Y`. That's exactly why `effectiveBackgroundColor`/`resolvedColors` exist.

So on a split theme the baseline can be the wrong side, and a TUI whose background matches the *dark* configured background won't be suppressed. Compare against the resolved color instead.

Also: no test covers this suppression threshold.

Macterm/Ghostty/AdaptiveTerminalBackground.swift
```diff
@@ -0,0 +1,186 @@
...
+            let key = Int(pixel.red >> 4) << 8 | Int(pixel.green >> 4) << 4 | Int(pixel.blue >> 4)
+            var bucket = buckets[key, default: Bucket()]
+            bucket.count += 1
+            bucket.red += Int(pixel.red)
+            bucket.green += Int(pixel.green)
+            bucket.blue += Int(pixel.blue)
+            buckets[key] = bucket
+        }
+
+        guard let winner = buckets.values.max(by: { $0.count < $1.count }) else { return nil }
```
thdxg, line 108: **The returned color can be one no pixel actually had.**

Bucketing is 4 bits per component, so a bucket spans a 16×16×16 RGB range, and the result is the *mean* of everything in the winning bucket. Near a bucket boundary that mean can sit visually outside the cluster it came from. `nearbyRendererValuesShareAQuantizedBucket` covers the benign case, and the 60% coverage floor limits the damage, but a gradient background that happens to concentrate in one bucket yields a tint matching nothing on screen.

The bucket's modal or median pixel would be more faithful than the mean.

Macterm/Ghostty/AdaptiveTerminalBackground.swift
```diff
@@ -0,0 +1,186 @@
...
+        let coverage = Double(winner.count) / Double(pixels.count)
+        guard coverage >= minimumCoverage else { return nil }
+
+        return Match(
+            red: UInt8(winner.red / winner.count),
```
thdxg: **Trapping conversion guarded only by a non-obvious invariant.**

`winner.red / winner.count` is `Int` division, so it truncates before the `UInt8` conversion (minor, ±1). The bigger issue is that `UInt8(_:)` traps on out-of-range. It can't overflow today — every summand is ≤255 and the divisor is the count — but that's a subtle invariant protecting a crashing conversion in a path that reads renderer memory.

`UInt8(clamping:)` costs nothing and removes the reasoning burden entirely.

Macterm/Ghostty/AdaptiveTerminalBackground.swift
```diff
@@ -0,0 +1,186 @@
...
+enum AdaptiveTerminalBackgroundPresentation {
+    static func windowColor(for paneColors: [NSColor?]) -> NSColor? {
+        paneColors.count == 1 ? paneColors[0] : nil
+    }
+
+    /// Every detected color also fills its own pane opaquely — including a
+    /// lone pane. The window-wide tint follows the user's window opacity, so
+    /// on a translucent window it alone would leave a visible seam between
+    /// the TUI's opaque pixels and the pane padding around them.
+    static func paneColors(for paneColors: [NSColor?]) -> [NSColor?] {
```
thdxg: **This abstraction is an identity function.**

`paneColors(for:)` returns its argument unchanged, and `windowColor` is a one-line ternary. Two of the new tests assert the behavior of an identity function, which doesn't buy much.

The doc comment explaining *why* pane colors pass through unchanged (the opaque-fill-vs-translucent-window seam) is genuinely valuable — I'd fold both functions into the call site in `refreshPresentation` and keep that comment there.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
```diff
@@ -0,0 +1,249 @@
...
+/// Samples every visible pane in the active terminal window and publishes its
+/// temporary terminal-app background. A single pane may tint the whole window;
+/// split panes are always isolated to their own bounds. Sampling follows render
+/// activity plus a low-frequency monitor for static TUIs and costs nothing
+/// until the user enables the feature.
+@MainActor
+final class AdaptiveTerminalChrome {
+    static let shared = AdaptiveTerminalChrome()
+
+    private var stabilizers: [ObjectIdentifier: AdaptiveTerminalBackgroundStabilizer] = [:]
```
thdxg: **`ObjectIdentifier` keys can be reused after dealloc.**

Views are weakly held in `liveViews`, so a view can dealloc and a fresh one land on the same address, matching a stale stabilizer. `pruneState` runs every sample so the window is small, but the pane's `UUID` is stable, already threaded through this code, and would remove the hazard entirely.

Macterm/Settings/SettingsView.swift
```diff
@@ -250,9 +250,21 @@ private struct AppearanceSettings: View {
     private var liquidGlassStyle: WindowGlassStyle = Preferences.shared.windowGlassStyle
     @State
     private var paneDimOpacity: Double = Preferences.shared.paneDimOpacity
+    @State
+    private var adaptiveTerminalChrome: Bool = Preferences.shared.adaptiveTerminalChromeEnabled
 
     var body: some View {
         Form {
+            Section("Terminal") {
```
thdxg: Placing this above `Section("Window")` makes an off-by-default experimental feature the first thing in Appearance settings. Suggest moving it below Window.

The label also reads as an OS full-screen-mode setting rather than "full-screen TUI" — something like "Match terminal app backgrounds" is less ambiguous.

Macterm/Settings/SettingsView.swift
```diff
@@ -250,9 +250,21 @@ private struct AppearanceSettings: View {
...
+                Toggle("Match full-screen app backgrounds", isOn: $adaptiveTerminalChrome)
+                    .onChange(of: adaptiveTerminalChrome) { _, enabled in
+                        Preferences.shared.adaptiveTerminalChromeEnabled = enabled
+                    }
+                Text("Matches the whole window for a single pane; in a split, only each full-screen app's pane changes color.")
+                    .font(.system(size: 11))
```
thdxg: Hardcoded font size — worth checking whether the other Settings descriptions use `.caption`/`.footnote`, and matching them.

thdxg (changes requested):
@OpenSource03 Thanks! I've left some comments on the diff.

thdxg:
@OpenSource03 I can fix conflicts and address the comments I made if you're okay with it!

OpenSource03:
> @OpenSource03 I can fix conflicts and address the comments I made if you're okay with it!

Hey! Sorry, I was on vacation for the past time, so I did not have time to go properly through those. I will handle your comments tomorrow or the day after.

If you want to go over it yourself and ship earlier, feel free to let me know, it also works for me.

thdxg:
@OpenSource03 No worries! No pressure on hurrying up. I'll leave it up to you!

Macterm/Ghostty/GhosttyApp.swift
OpenSource03: `MactermTheme.nsBg` reads `effectiveBackgroundColor`, which reads the observable `adaptiveBackgroundColor`. Observation tracks that dependency through computed accessors, including the static theme accessor.

I added a `withObservationTracking` regression test for this path. I’m keeping `configVersion` unchanged because the adaptive tint is transient window state, not a Ghostty configuration or theme-cache change.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
OpenSource03: Yep, I added the format check you suggested, plus a couple of bounds checks so we just stop sampling if Ghostty changes the layout. There’s a test for it now, and I noted the limitation in AGENTS.md.

I’d still keep sampling, though. OSC 11 only helps when an app changes the terminal background setting. Some TUIs just paint the whole screen a color, so OSC 11 alone misses the case this feature is meant to handle.

Macterm/Ghostty/AdaptiveTerminalChrome.swift
OpenSource03: I checked this path again. `backgroundColor` first uses the live surface’s `resolvedColors`, then the appearance-resolved theme file, and only then falls back to the app-global getter. It therefore already uses the active dark side of a split theme when available.

`effectiveBackgroundColor` is not the right baseline because it includes the adaptive tint itself, making the comparison self-referential. I kept the resolved configured background and added coverage for the suppression threshold.

OpenSource03:
> @OpenSource03 No worries! No pressure on hurrying up. I'll leave it up to you!

All fixed, apart from a few comments I have replied to. Finally 😅

thdxg:
@OpenSource03 I'm testing this locally, and the background doesn't seem to refresh immediately when I open a TUI with bg (e.g. opencode). It only refreshes when I move focus to another pane or tab. Can you look int this?

Macterm/Ghostty/AdaptiveTerminalChrome.swift
OpenSource03: Removing the repeating timer exposed a startup race: we could sample before Ghostty had drawn the TUI, then never try again until a tab switch.

I fixed that by using the existing output heartbeat and starting a short, bounded retry burst after output/render/focus events. This catches the finished frame without bringing back idle polling.

OpenSource03:
> @OpenSource03 I'm testing this locally, and the background doesn't seem to refresh immediately when I open a TUI with bg (e.g. opencode). It only refreshes when I move focus to another pane or tab. Can you look int this?

Yes, there was a regression with the timer after last fixes. I have, by mistake, retested it on an older version. I've confirmed it works now.

thdxg:
@OpenSource03 Thanks for your work! Looks good to me!

PR #185, opened 2026-07-24
fix: detect in-place terminal activity reliably

onliner10:
## What

Makes the tab activity indicator follow real terminal work reliably, including in-place redraws from same-PID Pi `!` commands and output from fully occluded tabs. Active tabs now render their actual execution state without special masking.

Also exposes the underlying `idle` / `running` / `done` state additively in `macterm pane list`.

## Why

The existing renderer/scrollback signals stop when a surface is occluded and miss raw-mode TUIs that redraw in place. Process polling cannot identify nested Pi shell commands because Pi keeps the same foreground PID throughout the command.

## How

- Consume Ghostty's new throttled `GHOSTTY_ACTION_OUTPUT_ACTIVITY` signal from the PTY IO path.
- Preserve explicit progress and canonical foreground processes as authoritative execution sources.
- For ambiguous same-PID agents, require two timely heartbeats after an actually forwarded nonempty submission; blank Return, startup output, typing, IME composition, and ordinary raw TUIs do not start activity.
- Track committed input across physical Return/keypad Enter, IME commits, paste resolution, and control-CLI `sendText` / `sendKey` paths, with destructive edits and cancellation clearing stale evidence.
- Quiet-settle activity-owned runs through a dedicated deadline that still polls when ordinary polling is coalesced or the window is fully occluded.
- Reject or refresh cached GhosttyKit artifacts that lack the required ABI.
- Keep the new control-protocol state optional so older responses remain decodable.

## Verified

Validated against the official `thdxg/ghostty` `build-2026-07-24` universal framework, built from `c459b2a` after thdxg/ghostty#5:

- Correctly submitted Pi `! sleep 6` transitioned `running → done` while the window was minimized, with the foreground Pi PID unchanged.
- Physical Return and keypad Enter, plus programmatic `sendText` / `sendKey`, start and finish activity correctly.
- Physical and programmatic blank Return stay idle; Return in Vim stays idle.
- Canonical foreground commands remain running through silent periods.
- Output after completion does not resurrect a finished state.
- Release archive and DMG build successfully.

## Notes for reviewers

The Ghostty dependency is already merged and released: thdxg/ghostty#5. The output heartbeat is emitted from the PTY IO path rather than the renderer, so it remains available while the terminal surface is occluded.

The local Xcode installation prints its existing stale-CoreSimulator warning, but macOS tests and the release archive complete successfully.

## Summary by CodeRabbit

* **New Features**
  * Pane listings and tab status indicators now expose execution state: `idle`, `running`, `done` (including when tracking is enabled).
* **Bug Fixes**
  * Improved terminal “quiet settle” behavior to be consistent regardless of pane occlusion.
  * More robust Return/command submission detection and clipboard paste handling (including safer clipboard parsing).
* **Documentation**
  * Updated setup guidance for GhosttyKit compatibility, refined tab naming/activity notes, and expanded `pane list` output documentation to include execution state.


scripts/setup.sh
```diff
@@ -14,11 +14,23 @@ RESOURCES_MARKER="Macterm/Resources/terminfo"
...
 need_xcframework=true
 need_resources=true
 need_zmx=true
-[[ -d "$XCFRAMEWORK_DIR" ]] && need_xcframework=false
+if [[ -d "$XCFRAMEWORK_DIR" ]]; then
+  if has_output_activity_action; then
+    need_xcframework=false
+  else
+    echo "Existing GhosttyKit lacks GHOSTTY_ACTION_OUTPUT_ACTIVITY; refreshing it"
+    rm -rf "$XCFRAMEWORK_DIR"
```
thdxg: This `rm -rf` runs before the replacement is known-good. If the latest release turns out to lack the ABI, the `exit 1` down at line 55 fires *after* the working (if older) framework is already gone — so a tree that could previously build is left with no framework at all, and re-running setup hits the same failure. Worth downloading and validating into a temp location, then swapping, so a bad release is a no-op rather than a regression.

Separately, `GHOSTTY_HEADER` hardcodes the `macos-arm64_x86_64` slice name. If the fork ever ships a differently-named slice, `has_output_activity_action` silently returns false and you get a re-download every single setup run with no indication why. A glob over the slice dirs would be more robust.

scripts/setup.sh
```diff
@@ -37,6 +49,11 @@ if $need_xcframework; then
   gh release download "$LATEST_TAG" --pattern "GhosttyKit.xcframework.tar.gz" --repo "$FORK_REPO"
   tar xzf GhosttyKit.xcframework.tar.gz
   rm GhosttyKit.xcframework.tar.gz
+  if ! has_output_activity_action; then
+    echo "Error: GhosttyKit from $LATEST_TAG lacks GHOSTTY_ACTION_OUTPUT_ACTIVITY" >&2
+    echo "The thdxg/ghostty output-activity downstream patch must be released first." >&2
+    exit 1
```
thdxg, line 78: Worth calling out that this makes the app unbuildable against any GhosttyKit release predating the fork patch — including when bisecting across this commit. That's a defensible call given the feature genuinely requires the ABI, but the error message could say so directly ("checkouts before <this commit> need an older GhosttyKit; see AGENTS.md") to save the next person the archaeology.

Macterm/Model/SplitNode.swift
```diff
@@ -725,7 +913,8 @@ final class Pane: Identifiable {
         sessionName persistedSessionName: String? = nil,
         command: String? = nil,
         shell: String? = nil,
-        env: [String: String]? = nil
+        env: [String: String]? = nil,
+        activityQuietPollDelay: Duration = .seconds(3)
```
thdxg: `activityQuietPollDelay` defaults to 3s and `settleTerminalActivityIfQuiet` defaults `quietInterval` to 3s (line 591). So the wake fires at almost exactly the threshold it needs to have crossed: the settle requires `now - lastActivityAt >= 3`, and the sleep targets 3.0s after the last heartbeat.

Today this works because `Task.sleep` jitter is positive. But it's load-bearing on that assumption, and the failure mode is bad: if the wake ever lands marginally early the settle no-ops, and with every window occluded there's no timer left to retry — the run stays `.running` indefinitely, which is precisely what this machinery exists to prevent.

Suggest making the margin explicit, e.g. delay = quietInterval + 0.25s, ideally deriving both from one constant so they can't drift apart.

Macterm/Model/SplitNode.swift
```diff
@@ -459,10 +594,47 @@ final class Pane: Identifiable {
...
+    /// ABI that emits it, so the quiet-settle needs no occluded-pane exemption.
+    /// Growth-vs-keepalive is decided in `TerminalExecutionTracker`.
+    func markOutputActivity(totalRows: UInt64, now: Date = Date()) {
+        executionState = executionTracker.markOutputActivity(totalRows: totalRows, at: now, currentState: executionState)
+        scheduleActivityQuietPollIfNeeded()
     }
 
-    func refreshTerminalActivityWindow(now: Date = Date()) {
-        executionTracker.refreshActivityWindow(now: now)
+    private func scheduleActivityQuietPollIfNeeded() {
```
thdxg, line 641: Minor: this cancels and recreates a `Task` on every `markOutputActivity`, i.e. ~2 Hz per pane with live output. Bounded by the throttle so it's not a real cost, but `DispatchWorkItem` would be lighter and matches the idiom already used a few files over for `commandSubmissionEvidenceReset`.

Macterm/Views/Terminal/GhosttyTerminalNSView.swift
```diff
@@ -1161,12 +1254,18 @@ extension GhosttyTerminalNSView {
         // Same liveness signal a keystroke sends (execution tracking + poll
         // resume), so an injected command updates the tab title promptly.
         onInteraction?()
+        recordCommandInput(text)
```
thdxg, line 1277: `recordCommandInput(text)` records the same string that's checked for a newline two lines down, so `hasContent` here is derived from the caller's own payload rather than from prior typing. Fine for `pane run "sleep 5"`.

The issue is that `Evidence.recordText` only ORs in — it never clears — combined with `preserveProgrammaticCommandInput`'s 2s window. So a `pane run` followed within 2s by a genuinely blank Return will report `hasContent == true` for that blank Return, which is the exact false positive the blank-submission suppression exists to prevent. Narrow, but reachable from a script that injects a command and then sends a bare newline.

Macterm/Model/SplitNode.swift
```diff
@@ -80,46 +87,92 @@ struct TerminalExecutionTracker {
...
-    /// render/output heartbeat and quiet-settles.
+    /// progress run until a completion/foreground transition; activity is an
+    /// output heartbeat and quiet-settles.
     private var runningSource: TerminalExecutionSource?
     /// After progress clears, the foreground process that owned it is
     /// "quiesced": its own output and re-polls are ignored until the foreground
-    /// moves away, so a settled program that reported progress doesn't flip back
-    /// to running on its own render output. `pendingProgressQuiesce` covers the
-    /// race where progress started and cleared before any foreground poll.
+    /// moves away. `pendingProgressQuiesce` covers progress that starts and
```
thdxg, line 120: `recordUserInteraction` now clears `pendingOutputStart`, which makes the arming call sites order-dependent: `keyDown`, `sendText`, and `sendKey` all fire `onInteraction?()` before `onCommandSubmitted?()`, so it's clear-then-arm and works. Reorder either call and the Pi path silently stops arming, with no test to catch it.

I tried to construct a failing case and couldn't, so this isn't a bug report — just asking for a line of comment here stating that callers must arm *after* interaction, since the constraint is currently invisible at both ends.

Macterm/Views/Terminal/TerminalCommandSubmission.swift
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    static func clearsInputEvidence(
+        keyCode: UInt16,
+        hasControl: Bool,
+        hasCommand: Bool
+    ) -> Bool {
+        if keyCode == 53 || keyCode == 51 || keyCode == 117 { return true } // Escape / delete
+        if hasControl, [4, 8, 13, 32, 40].contains(keyCode) { return true } // H/C/W/U/K
```
thdxg: The keycodes check out (4=h, 8=c, 13=w, 32=u, 40=k), but bare literals with a `// H/C/W/U/K` gloss are easy to get wrong on edit. `HotkeyRegistry.parseShortcut` already carries a keycode vocabulary the CLI leans on — reusing it, or at minimum naming these, would be more maintainable.

Also line 54: `// Escape / delete` is two labels for three codes (53 escape, 51 delete, 117 forward-delete).

MactermTests/Model/PaneTests.swift
```diff
@@ -287,16 +297,76 @@ struct PaneTests {
...
+                forName: .terminalQuietSettleDeadline,
+                object: p,
+                queue: .main
+            ) { _ in
+                confirm()
+            }
+            defer { NotificationCenter.default.removeObserver(token) }
+
+            p.markOutputActivity(totalRows: 20)
+            try? await Task.sleep(for: .milliseconds(100))
```
thdxg: Fixed 100ms sleep against a 20ms delay — this is the shape that's flaked on loaded CI runners before (cf. #181). Better to poll until the confirmation fires with a sleep in the loop, or widen the margin substantially, so a co-scheduled spike doesn't fail the run.

Macterm/Ghostty/GhosttyCallbacks.swift
```diff
@@ -117,8 +121,14 @@ final class GhosttyCallbacks: @unchecked Sendable {
         // `ghostty_surface_complete_clipboard_request`'s surface parameter is
         // non-null on the Zig side; a request completing during surface
         // teardown (nil `surface`) is UB, not a graceful no-op. Guard it.
-        guard let surface = surface(from: ud) else { return false }
+        guard let ud else { return false }
```
thdxg: Functionally identical to the `surface(from:)` helper this replaces, so the teardown guard the comment describes is preserved — but it open-codes `Unmanaged.fromOpaque` at a second call site, and the comment above now refers to a guard spread across three lines. A `surfaceView(from ud:)` overload would keep the pointer-casting in one place.

Also worth a note in the comment: `surfaceDidPasteText` is dispatched async *before* the synchronous `complete_clipboard_request`, so the evidence lands after the paste reaches the surface but before any subsequent Return. That ordering is correct and load-bearing, and currently unremarked.

thdxg (changes requested):
@onliner10 Thanks! I've left some comments

scripts/setup.sh
onliner10: Both fixed. Download+validate now happens in a `mktemp -d`, and the old framework is only removed once the replacement passes, so a bad release is a no-op. Probe takes the root as an arg and globs `*/Headers/ghostty.h` instead of naming the slice.

scripts/setup.sh
onliner10: Good call. The error now says the ABI has been required since this change, that a pre-change checkout needs a GhosttyKit from a contemporary release, and that the existing one was left alone. Expanded the AGENTS.md paragraph it points at.

Macterm/Model/SplitNode.swift
onliner10: Yep, that was relying on jitter sign. Added `TerminalActivityTiming` with `quietInterval = 3` / `quietPollMargin = 0.25` / `quietPollDelay = interval + margin`; the wake and the settle default both derive from it, and a test pins the margin so it cannot collapse.

Macterm/Model/SplitNode.swift
onliner10: Switched to `DispatchWorkItem` + `asyncAfter`. `activityQuietPollDelay` is a `TimeInterval` now.

Macterm/Views/Terminal/GhosttyTerminalNSView.swift
onliner10: Reachable, yeah. The view cannot tell the two cases apart (it does not know the payload already executed), so I scoped the carry rather than tried to detect it: it is gated on `Pane.allowsInPlaceOutputStart` now, which is the same agent-TUI foreground the in-place heuristic needs and the only place a bracketed paste can eat the newline. `pane run` into a plain shell no longer opens the window.

Macterm/Model/SplitNode.swift
onliner10: Documented at both ends: an ordering note on `recordUserInteraction`, one on `onCommandSubmitted`, and one at the `TerminalPane` wiring.

Macterm/Views/Terminal/TerminalCommandSubmission.swift
onliner10: Named them, and split the escape/delete gloss into three cases. Cannot read `HotkeyRegistry` at runtime here since it is `@MainActor` and this helper is isolation-free, so I added `keyCode(forToken:)` and a test that drives the codes from that vocabulary through the public API.

MactermTests/Model/PaneTests.swift
onliner10: Fixed. Polls in 25ms ticks with a 5s ceiling, so a healthy run exits on the first tick and a spike cannot fail it.

Macterm/Ghostty/GhosttyCallbacks.swift
onliner10: Added a `surfaceView(from ud:)` overload with `surface(from:)` delegating to it, so the cast lives in one place. Also noted why the paste record is dispatched before the synchronous completion.

thdxg (approved):
All nine addressed. I verified each against the code rather than the replies, and built and ran the branch this time.

**The two blockers are genuinely fixed:**

`setup.sh` now stages into `mktemp -d`, validates there, and only `rm -rf`s the existing framework once the replacement passes — with a `trap` for cleanup. I got to exercise this for real: my checkout's GhosttyKit predated the ABI, so `mise run setup` took exactly the path in question. It refreshed correctly, swapped in a framework with the ABI present, and left no staging dir behind. The slice glob replaces the hardcoded name as asked.

`TerminalActivityTiming` is the right shape — one constant, `quietPollDelay = quietInterval + quietPollMargin`, with both the wake and the settle default deriving from it and `quietPollWakeLandsAfterTheSettleThreshold` pinning the relationship so it can't collapse back.

**On the `sendText` carry (my #1):** scoping to `allowsInPlaceOutputStart` is a better answer than what I proposed. I was thinking about detecting the two cases; you correctly noted the view can't, and instead narrowed the carry to the only foreground where a bracketed paste can actually swallow the newline. `pane run` into a plain shell no longer opens the window at all, which removes the false positive rather than papering over it. The shared property also keeps the gate and the heuristic from drifting.

**On the keycodes:** `keyCode(forToken:)` plus `keyCodesMatchHotkeyRegistry` is a cleaner resolution than reading the registry at runtime would have been — you kept `TerminalCommandSubmission` isolation-free and still got the drift protection. Good call, and thanks for explaining the `@MainActor` constraint rather than just working around it.

The ordering contract is now documented at all three ends, `GhosttyCallbacks` has the single cast site plus the async-before-completion rationale, and the flaky sleep is a poll with a ceiling.

**Verified locally** on `8c35511`: `mise run test --verbose` passes (`PaneTests`, `TerminalCommandSubmissionTests`, `TerminalExecutionTrackerTests` all green), `mise run lint` clean, `mise run format` reports 0/125 reformatted.

One leftover from my first pass, for the record: I flagged `return true` on the nil-view path in `GhosttyCallbacks` as inconsistent. That was wrong — it's what every case in that switch does. No action needed; noting it so it doesn't get "fixed" later on my say-so.

The `displayState` removal is still a user-visible change that isn't in the PR description — the active tab can now show a checkmark between a command finishing while the app is inactive and the next poll acknowledging it. Not blocking, and arguably more honest than masking, but worth a line in the body before squashing since the squash subject is what lands in history.

thdxg:
@onliner10 Thanks for your work!

PR #189, opened 2026-07-27
Add an end-to-end CI suite that drives the real app over the control socket

### LINKED ISSUE #121: Add integration and end-to-end testing layers
## The gap

Our unit coverage is strong (~340 tests over model, persistence, palette, hotkey, and config logic), and not unit-testing SwiftUI/AppKit/libghostty is the right call. But that leaves a whole category untested: we have no integration or end-to-end tests at all, and CI runs only format, lint, and unit tests. Nothing checks that the pieces work once assembled, and nothing checks that a real terminal session behaves correctly from the first keypress. That's exactly where our worst regressions came from — the broken `TERM=xterm-ghostty` and dead keyboard input in #39/#40 shipped because no test could catch them.

Two tiers would close this. **Integration tests** check that the assembled app is wired correctly; **end-to-end tests** check that a real session actually works, with nothing stubbed (real shells, PTYs, libghostty, and relaunch cycles).

## Integration tier

Drive the built app with XCUITest, seed known state before launch, and assert on the result. Seeding is easy: the app already reads `-macterm.activeProjectID` and persists to a per-build data dir (`Macterm Debug/`), so a test can drop in a known `projects.json`/`workspaces_v3.json`/`.macterm/layout.yaml`, launch, assert, and clean up. The high-value flows are workspace restore on launch, declarative layout auto-apply on first open, the config pipeline actually applying (our overrides winning over the user's Ghostty config), single-window enforcement, and focus landing correctly after tab/split reshapes — all things unit tests prove in isolation but never prove assembled.

## End-to-end tier

This runs real terminal sessions and asserts on terminal behavior, not app state. The one wrinkle: the surface is Metal-rendered, so its text isn't readable via accessibility. The fix is to assert through side channels the shell can write — have the shell run `echo OK > "$OUT"` (or `printf '%s' "$TERM"`, `infocmp xterm-ghostty`, etc.) and poll the file. That's actually a better test of "did the command really run" than reading pixels, and it needs no special privileges.

The scenarios that matter most are a keystroke-to-execution round-trip (the basic "the terminal works," and a guard against #40), `TERM`/terminfo correctness with a real TUI launching and exiting cleanly (the #39/#40 breakage, ideally run once on a machine with no Ghostty.app so bundled resources are truly exercised), shell integration propagating cwd/title, and a full session surviving quit-and-relaunch. Clipboard, scrollback (#102), and resize/reflow are worth adding later.

## CI

Both tiers need a GUI login session. Hosted `macos` runners can run basic XCUITest and most of the side-channel E2E tests; anything needing Accessibility or dismissing system dialogs can't run there. So keep both out of the required PR check and tier execution: local always, a nightly hosted job for the grant-free subset, and a self-hosted runner with Accessibility pre-granted for the rest if it earns its keep.

## Scope

Additive only — this doesn't change the unit suite or the "don't unit-test the UI" rule, and it's not about snapshot/visual tests or libghostty internals. A reasonable start is the harness plus two smoke tests (workspace restore + keystroke round-trip), expanding from there.

thdxg:
## What

End-to-end GUI testing in CI, built on infrastructure the repo already had instead of XCUITest:

- **`scripts/_harness.py`** — the benchmark's launch/drive/teardown recipe (LaunchServices launch, Darwin-notification control, hermetic env, control-socket CLI), extracted into a `MactermHarness` shared by `benchmark.py` and the new suite. Verified both ways: the refactored benchmark ran a full workload smoke green, and the suite uses the same code paths.
- **`e2e/`** — 11 pytest tests against a live, hermetically launched app: boot smoke (project open → live shell → readable prompt → foreground poll), terminal I/O round-trips (a typed command's *output* found via `pane dump`; ctrl+c through `pane key` actually interrupting a process), and tab/pane semantics (split, grid, focus, idle close, busy-close refusal + `--force`).
- **`mise run e2e`** (`scripts/e2e.sh`) — Debug build + venv-bootstrapped pytest; Debug on purpose (correctness is under test, shares the unit-test job's DerivedData cache, keeps debug-only verbs available for later reflow tests).
- **`e2e` job in checks.yml** — separate from `test` so the slower GUI run never delays the fast unit signal; failures upload a diagnostics artifact (screenshot, streamed unified log, per-pane terminal text).

## Why not XCUITest

The terminal is a `CAMetalLayer` — no accessibility tree, so XCUITest could click chrome but never assert "the shell printed X". The control socket can, with zero TCC grants, on a stock runner. A passing I/O test transits socket → AppState → libghostty → pty → shell → renderer cells and back.

## Latent bugs fixed in the extraction

- **Hermetic instances shared the user's zmx namespace.** The zmx socket dir is per-*user* (`TMPDIR/zmx-<uid>`), not `$HOME`-scoped, so a benchmark run's orphan reaper could kill the developer's real detached sessions, and its workload shells leaked (a `date` loop per pane, left running after every local `mise run bench`). The harness now sets `ZMX_DIR` into the throwaway home and tears down its own sessions, scoped by pane attachment — never by name.
- **Failure diagnostics never contained any app log.** `log show --level debug` isn't a flag (`--level` belongs to `log stream`), and `.info`/`.debug` messages don't persist to the datastore anyway. The harness now runs a live `log stream` for the app's lifetime.
- **Pid discovery could adopt a stray instance** (leftover from an aborted run, or the developer's own copy of the same build). Only pids that appear after this launch count now.

## Reviewer notes

- **The one flake found, and the rule it produced:** syncing a test on the foreground-*name* poll flakes — a fresh zmx session that misses the resolver's registration retry window waits for `ZmxRefreshGate`'s 30s reconcile TTL (diagnosed with an instrumented probe on a loaded machine; benign in the product, the tab title just shows the shell-name fallback meanwhile). The first CI run then disproved the obvious replacement too: `needsConfirmQuit` is only accurate where shell integration reaches the zmx session shell — ghostty has no integration for bash 3.2, which is what the CI runner's user logs into, so there every pane reads perpetually busy. Tests therefore prove execution state with runtime-assembled output markers (`printf started-%s <nonce>` — the typed/echoed line never contains the assembled string, so a dump match proves execution, not keystroke echo; ctrl+c delivery = an `after` marker executes while `finished` never prints), and the close-without-force contract is asserted as what it really is: close follows the `needsConfirmQuit` signal, whichever way the environment sets it. Every wait is a deadline poll; there are no fixed sleeps and no retry-on-failure.
- **`pane run` footgun** (documented in the harness): its passthrough capture swallows trailing flags, so `--socket`-last would fall back to socket discovery and could type into a *real* Macterm. `MactermHarness.pane_run` places flags before the command.
- Commands typed into panes are `/bin/sh -c "…"` with single-quote-free scripts so they parse under any login shell (CI's runner user logs into bash 3.2; dev machines may run nushell) — same trick as the benchmark workload and `RemoteSpawn`.
- `benchmark.yml`'s path filter includes `scripts/**`, so this PR's benchmark run exercises the refactored harness in CI too.
- Local result: 11/11 in ~11s of test time (~30s wall including the one app launch). Suite conventions are documented in CLAUDE.md.

## Follow-ups (not in this PR)

Phase 2 of the plan: persistence/restart test (kill → relaunch same data dir → same `sessionName`, scrollback survives), layout apply/save round-trips, scrollback/reflow guard using the debug-only `pane resize` (#112 class), zoom/resize-split. Phase 3: loopback-ssh remote project test (nightly), a debug-only `command invoke <AppCommand-id>` verb for palette-action coverage.


thdxg:
A note on the `benchmark:improvement` label: treat it as a measurement artifact, not a real win from this PR.

- This PR changes no app code — `Macterm/`, `CLI/`, and `project.yml` are untouched vs main; only scripts, tests, workflows, and docs changed. There is no mechanism for it to cut app CPU.
- The two benchmark runs of this same branch disagree with each other more than the claimed effect: `workload-unfocused` CPU read **−36%** on the first run and **+14%** on the second. Both compare against one main baseline sample (`main@f1cd144`), so what the label is detecting is runner-to-runner variance — a faster bare-metal runner moves *all* cells in the same direction, which the ≥2-cells corroboration rule can't distinguish from a genuine uniform improvement.
- This self-heals after merge: the first `main` run rebaselines with the same harness, so subsequent PRs compare like-to-like.

Worth keeping in mind as a known gap in the labeling guards: corroboration defends against single-cell noise, but not against whole-runner shifts. A cheap future mitigation could be comparing against the median of the last N main baselines instead of the single latest.

jcuhnoio:
Love to see this happen!!

PR #199, opened 2026-08-02
Let Apply/Save Layout be bound to a key

### LINKED ISSUE #113: Some MacTerm UI Improvements

thdxg:
## What

Makes **Apply Layout** and **Save Layout** rebindable commands. Both ship unbound (`none`) — they gain a Settings → Keymaps row and a working binding, not a default shortcut.

## Why

Both were palette-only. They're the kind of action you reach for repeatedly while iterating on a project's pane layout, and every comparable command is already bindable.

Unbound by default is deliberate rather than just conservative: Apply Layout rewrites the live pane tree and Save Layout overwrites the project file, so a stock chord would make a mistyped binding destructive. Opt-in via Settings.

## How

Follows the "Adding a New Action" recipe in CLAUDE.md:

1. `HotkeyAction` gains `applyLayout` / `saveLayout` (`apply_layout` / `save_layout`), both defaulting to `"none"`.
2. `AppCommand.hotkeyAction` links them — they move out of the `nil` group. Titles flow from `AppCommand`, so Settings renders "Apply Layout"/"Save Layout" under Projects with no new strings.
3. `MainAppResponder` appends both to the existing loop that dispatches through `AppCommand.action(in:)`.

Step 3 is what keeps this small. Neither command gets a hand-written handler, so both inherit the palette's enablement guards — Apply Layout with no applicable project file yields a `nil` action and the keystroke falls through instead of firing a no-op — plus the same error-presenting wrappers (`applyLayoutPresentingError` / `saveLayoutPresentingError`). The key path and the palette path can't drift.

Settings → Keymaps picks both up automatically via `HotkeyAction.allCases`, as does `isAppShortcut` in `GhosttyTerminalNSView`, so once bound the chord passes through the terminal instead of being typed into the shell.

## Verified

`HotkeysTests.layout_actions_are_unbound_by_default_and_titled_from_command` pins the unbound defaults and the command↔action links, mirroring the existing `copySessionID` test. Confirmed it actually ran and passed in the xcresult rather than relying on a green suite.

Not exercised in a running app — the change is config/dispatch wiring with no default binding to press, and the dispatch path it joins is already covered by the three commands using it.

## Notes for reviewers

Apply/Save Layout have **no menu-bar items** today (palette-only), and I left that alone as outside the ask — so they gain a Settings row and a binding but no Project-menu entry. Adding them there is a reasonable follow-up.

Worth a look: the two new `HotkeyAction` raw values are persistence keys (`macterm.hotkey.apply_layout` / `…save_layout`), so renaming a case later is a stored-preference migration.

### DIFF
--- a/Macterm/App/AppCommand.swift
+++ b/Macterm/App/AppCommand.swift
@@ -169,12 +169,12 @@ enum AppCommand: String, CaseIterable, Identifiable {
         case .renameTab: .renameTab
         case .renameProject: .renameProject
         case .copySessionID: .copySessionID
+        case .applyLayout: .applyLayout
+        case .saveLayout: .saveLayout
         case .newRemoteProject,
              .unloadProject,
              .removeProject,
              .replaceProjectPathWithCurrentDir,
-             .applyLayout,
-             .saveLayout,
              .checkForUpdate: nil
         }
     }
--- a/Macterm/App/Hotkeys.swift
+++ b/Macterm/App/Hotkeys.swift
@@ -39,6 +39,8 @@ enum HotkeyAction: String, CaseIterable, Identifiable {
     case renameTab = "rename_tab"
     case renameProject = "rename_project"
     case copySessionID = "copy_session_id"
+    case applyLayout = "apply_layout"
+    case saveLayout = "save_layout"
 
     var id: String { rawValue }
 
@@ -80,6 +82,10 @@ enum HotkeyAction: String, CaseIterable, Identifiable {
         case .renameTab: "cmd+r"
         case .renameProject: "none"
         case .copySessionID: "none"
+        // Unbound by default: both rewrite or replace the live pane tree, so a
+        // stray keystroke on a stock binding would be destructive.
+        case .applyLayout: "none"
+        case .saveLayout: "none"
         }
     }
 }
--- a/Macterm/App/Responders.swift
+++ b/Macterm/App/Responders.swift
@@ -290,8 +290,11 @@ final class MainAppResponder: KeyResponder {
         // truth shared with the palette and menu bar — so the paths can't drift.
         // Rename defers begin-editing a tick (see AppCommandActions) so the
         // sidebar row's TextField exists before it takes first responder;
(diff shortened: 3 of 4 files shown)


thdxg:
Investigated rather than pushed a fix — this looks like a false positive, and I want to lay out why so the guard can be improved if you agree.

**The diff cannot execute during the benchmark.** The whole production change is two `HotkeyAction` enum cases (both defaulting to `"none"`), two `hotkeyAction` links, and two extra elements in a dispatch array in `MainAppResponder.handle`. That code runs **only on a key event**. The harness types nothing — its workload is `/bin/sh` date loops spawned through the CLI — and both flagged cells are `workload-unfocused`, where the window isn't even key.

The one plausible mechanism was `HotkeyAction.allCases` growing by two, since `GhosttyTerminalNSView.isAppShortcut` scans it per keystroke. I checked every `allCases` site: the other three are in `SettingsView`. None is on a timer, in the render path, or in the pty path.

**The two flagged cells aren't independent.** `should_label` requires ≥2 cells so a lone outlier can't label — but CPU % and CPU ms/s are two instruments reading one quantity in one state. Their ratio is unchanged:

| | ms/s per CPU % |
|---|---:|
| baseline | 9.90 |
| this branch | 9.96 |

0.6% apart — one CPU reading counted twice. They also clear their floors by the same margin (0.70 vs 0.5 = 1.40×; 7.1 vs 5.0 = 1.42×), which is what one number in two units looks like, rather than two correlated effects.

Supporting signal: idle `focused` CPU moved +50% / +59% on the same run — a state this diff cannot touch at all — which reads as a hot runner across the whole run rather than a code change.

**Possible follow-up (not this PR):** `should_label` could count `cpu_pct` and `cpu_ms_per_s` within the same state as a single vote, so corroboration requires genuinely distinct signals or distinct states. Happy to open a separate PR if that seems right.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #206, opened 2026-08-02
Pin the GhosttyKit release, and bump it weekly by PR

thdxg:
`thdxg/ghostty` publishes a `build-YYYY-MM-DD` release **every day**, and `setup.sh` resolved `latest`. Every Macterm build — tagged releases included — therefore picked up whatever libghostty landed that morning. Two builds of the same Macterm commit could ship different terminal cores, and a release could carry an upstream change nobody had run.

This is the systemic issue behind the Sentry crash landing in a release unnoticed. `GHOSTTYKIT_TAG` now names the release, so bumping the terminal core is its own reviewable commit.

Pinned to `build-2026-08-02` — the first release built with `-Dsentry=false`.

## Design notes

- **One tag, both artifacts.** The xcframework and the resources tarball ship in the same release, and they refresh **together** — a bump must never leave a new libghostty beside the old release's terminfo.
- **A stamp, because presence checks aren't version checks.** Without it the existing checks would keep the old copy straight through a bump — the same silent-staleness trap that makes symlinking these artifacts a bad idea. `setup.sh` records the installed tag in `.ghosttykit-tag` (gitignored) and refreshes when it disagrees.
- **Stamp written last**, and only once both artifacts are on disk, so a setup that dies between the two downloads doesn't look complete to the next run.
- **No forced re-download on rollout.** A checkout predating the pin has no stamp, which is read as "matches"; it adopts one on its next real download.
- **CI needs no change.** All four cache keys already hash `setup.sh`, so bumping the pin changes the key and the next run downloads fresh. Verified across `test.yml`, `release.yml`, `benchmark.yml`.
- **Escape hatch:** `GHOSTTYKIT_TAG=latest mise run setup` (or any tag) to try one without committing.

## Verification

Ran setup through every path:

| Path | Result |
|---|---|
| Migration (artifacts present, no stamp) | early exit, no download |
| Pin bump (stamp mismatch) | refreshes both artifacts, stamp updated |
| Idempotent re-run (stamp matches) | early exit |
| `GHOSTTYKIT_TAG=latest` | resolves to `build-2026-08-02` |
| Nonexistent tag | **exits 1**, existing artifacts and old stamp left intact |

`bash -n` clean. `mise run test --verbose` passes on the pinned artifacts.

Note the `set -e` interaction on `[[ -f $TAG_STAMP ]] && stamped_tag=$(cat ...)`: exercised with the stamp absent (migration path) and it does not abort — and it matches the idiom already used for `RESOURCES_MARKER`/`ZMX_BIN`.


thdxg:
## Added: the bump path

The pin alone left nothing to move it off a rotting tag — "when do we update it?" had no answer. `bump-ghosttykit.yml` now runs **Mondays** (matching `dependabot.yml`'s actions group, same `dependencies`/`area:ci` labels and `chore(deps):` prefix), resolves the fork's latest release, and opens a `chore/ghosttykit-<tag>` PR when it differs from the pin. One branch per target tag, so a re-run while that PR is open is a no-op.

It uses **`GH_PAT`, not `GITHUB_TOKEN`** — a PR opened by the default token does not trigger workflows, which would land the bump *untested* and defeat the point of pinning.

**CI on the bump PR is the gate.** Editing `setup.sh` changes the GhosttyKit cache key, so the run downloads the new pin from scratch and puts it through Unit, End-to-end, and the benchmark. That is the check that would have caught the `sentry-init` crash before it shipped in a release — as this PR's own run demonstrates.

## Two things the pin exposed in `warn_if_ghosttykit_stale`

That warning is the backstop for when the scheduled runs fail silently, and pinning broke both halves of it:

1. **It was unreachable.** It sat after the download blocks, past setup's `already present` `exit 0`, so on a settled checkout it never ran. Tolerable while the tag was `latest` (any fresh checkout pulled current anyway); useless once nothing ever pulls current. It now runs on that path too — which meant hoisting it and `to_epoch` above the early exit.
2. **Its advice had become wrong.** It said `rm -rf GhosttyKit.xcframework && ./scripts/setup.sh`, which under a pin re-downloads *the same tag*. It now says to bump `GHOSTTYKIT_TAG` and names the currently pinned tag.

## Verification

- Pin parse against the real file → `build-2026-08-02`.
- `sed` rewrite hits **only** the assignment on line 19 — not the `GHOSTTYKIT_TAG=latest` usage comment above it, nor the resolved-tag `echo` further down — and the post-rewrite readback confirms the new value. File stays valid bash.
- Warning fires on the no-op path with the tag interpolated:

```
GhosttyKit, resources, and zmx already present

warning: bundled GhosttyKit (1.3.2-main-+d818a54) was built from an upstream base ~0 days
         Fix: bump GHOSTTYKIT_TAG in scripts/setup.sh (currently
         build-2026-08-02) to a newer thdxg/ghostty release, then re-run setup.
```

That version string is also independent confirmation the installed artifact came from `d818a54` — the `-Dsentry=false` merge.

## One caveat, now documented

The warning infers the upstream base as the fork commit's **parent**, which holds only while the fork's `main` is `upstream + 1 squash commit`. Landing a normal PR on the fork (as the `-Dsentry=false` change did) makes the parent a fork commit, so the reported age is meaningless until the next `sync-upstream` run rewrites `main`. Pre-existing, not introduced here, but worth knowing before trusting a `~0 days` reading.

Issue #227, opened 2026-08-06
Split pane by dragging existing tabs and separating panes


jcuhnoio:
### Description

Mainly inspired by Arc's split tabs, it would be useful if the user can drag tabs around and create split panes from existing tabs, and also be able to separate split panes into respective tabs.

Split panes will be created from existing tabs in two ways:
- Dragging a tab from the sidebar into another tab in the sidebar.
    - The user gets to choose where the tab goes in the horizontal axis for the initial split. If the "splitter tab" is dragged onto the left side of the "splitee tab", the splitter goes to the left, and vice versa.
    - This will be displayed in real-time in the sidebar, where the splitee tab moves around left or right.
    - This should only work for the first tab that gets split because once there are too many tabs in the split view, this would feel more janky than useful. After the first split, the splitee tab will be highlighted if a new splitter tab is dragged and hovered onto.
- Dragging a tab from the sidebar into the current pane.
    - The user gets to choose the direction of the split in both the horizontal and vertical axes as they have more screen space.
    - The current open tabs (the "splitee tabs") will move accordingly, showing a preview of where the tab will be dropped into before the user actually releases the tab.
    - The "splitter tab" will be a transparent window that follows the mouse that gets resized accordingly, and the space the splitter tab will go into would remain blank with an outline. 

Each tab in the split view will have a unique "container" rather than concatenating them with the pipe symbol, kinda like this:
<img width="217" height="42" alt="Image" src="https://github.com/user-attachments/assets/634efdaa-6b0d-42ed-b0c1-67c549c0e4dd" />

This will affect the sidebar UI of the current method of splitting panes as well.

Panes will separated when right-clicking the tab from the sidebar and selecting 'Separate panes', in which each pane in the split view will open into respective tabs.

### Motivation

I use this a lot with Arc when I have tabs that I have already opened that needs to be viewed simultaneously after they've been opened.

### Alternatives considered

I'm unsure of the technical feasibility of this as of now, like how much of a rewrite is needed to make this work with the current proposed UI. Simplifying UI animations while maintaining the intended feature may be the first step.

PR #228, opened 2026-08-06
Drag tabs into splits, rebuild pane reordering on one placement engine, and add Separate Panes

jcuhnoio:
Implements #227.

## What's here

- **Drag a sidebar tab onto another tab row** to merge the two into a split. On a single-pane destination, the half of the row you hover picks which side the dragged tab lands on — the splitee's row content slides aside in real time to show the drop slot (the issue's preview behavior); an already-split destination highlights whole and appends. The drag/drop region covers the entire row.
- **Drag a sidebar tab into the workspace, with level-aware targeting** (`TabDropPlacer`): the cursor picks not just a side but a LEVEL in the split tree, and the highlight previews exactly the region the tab will occupy.
  - Within 15% of a workspace edge on the layout's own axis → new top-level column/row, equalized (two columns become thirds). These side-by-side margins span the full cross-axis height, not gated by the corner triangles.
  - Within 15% of a divider → insert between the siblings, equalized (thirds).
  - In between → local split of the hovered pane (1/4 · 1/4 · 1/2).
  - On the perpendicular axis, the band from the midline to 75% takes the whole edge (e.g. a full-width bottom pane); deeper splits just the hovered pane.
- **The original pane reordering (the three-dots grab handle) is rebuilt on the same placement engine.** The old per-leaf drop targets could only reorder a pane relative to a single other pane; a grab-handle drag now resolves through the identical bands and preview, so a dragged pane can also take a whole edge or insert at a divider as an equal sibling. `TerminalTab.movePane(to:)` detaches the pane and re-merges it through the shared `mergeTree` — the moved pane always vanishes from its old spot (the tree collapses around it) and every other pane survives exactly once, pinned by test. The legacy `movePane(onto:zone:)` survives as a one-line alias into the shared path, so its whole test suite doubles as regression coverage; the quick terminal shares the same drop machinery (pane drags only).
- **Drop bands are capped per pane** at 30% of the hovered pane's extent, so skinny panes stay splittable: with four side-by-side columns the fixed workspace-relative bands used to swallow every pixel of every pane, making top/bottom placements unreachable. Two-pane layouts are numerically unchanged.
- **Separate Panes** context-menu item on split tabs: every pane after the first moves into its own tab, inserted in tree order after the source; shells and surfaces reused, never respawned.
- **Sidebar rendering for split tabs**: 2–3 panes render as one tab-like container per pane sharing the row (per-pane icon + title); 4+ panes collapse back to a single tab titled with the pane count ("4 panes"). A custom rename always wins, and clearing it falls back to the count.

## How

- The reshapes these drags produce remount the split tree, which exposed how fragile handing the pane-owned NSView directly to SwiftUI is (several distinct teardown orders can strand it). `TerminalSurface` now gives SwiftUI a disposable container per identity and re-parents the pane's real view into it, with layered healing — window-entry claim, orphan re-attach, late surface creation, and an observable `Pane.surfaceReattachTick` as the owner-of-last-resort so a stranded view always re-renders onto a container SwiftUI guarantees is alive. Verified with live `SurfaceAttach` event traces (debug logging kept in-tree) and headless repros via new DEBUG-only CLI verbs (`pane move`, `tab merge`) that drive the same code paths as the drags.
- `SplitNode.inserting(node:...)` generalizes the single-pane insert so a whole subtree lands beside a pane in one structural move; `tileUnits` is exposed to size drop previews.
- `TerminalTab.adoptTree`/`insertTree`/`mergeTree` are the tab-level merges — local drops keep the halving pref, divider/edge drops always equalize (their meaning is positional parity). `TabDropPlacer` is a pure resolver from a normalized cursor point to a target + preview rect.
- Workspace tab drops live on ONE workspace-level drop target (per-pane geometry can't see globality); pane drags keep the existing per-leaf zones. Both new drop delegates guard against the dropUpdated-after-performDrop race so highlights never stick.
- `AppState.mergeTab(intoTab:side:)` / `mergeTab(at:inProject:)` detach the source tab with panes alive, rebind routing identity on cross-project merges (mirroring `moveTab`), and restore the tab if the target vanished mid-drag. `separateTabPanes` is the inverse.

## Tests

- `TabDropPlacementTests`: the placement bands band-by-band, previews, full-height side-by-side margins, single-pane root, and nested-divider cases.
- `TerminalTabTests`: `adoptTree`/`insertTree`/`mergeTree` structure + ratio assertions (thirds, full-width half, 1/4-1/4-1/2), `sidebarRowTitle` rename behavior.
- `AppStateTests`: merge into tab (sides, whole-tree, self-merge no-op, cross-project rebind + activation), resolution-target merges, separate panes.


jcuhnoio:
Here's a demo:

https://github.com/user-attachments/assets/d0f2bb82-68f4-42f0-b55b-56ad8987ec6f

jcuhnoio:
I'd love to hear some feedback on the placement engine!

PR #229, opened 2026-08-07
Fix the hidden sidebar's hover reveal: full-width peek that retracts without the native overlay's broken titlebar

jcuhnoio:
## What's here

With the sidebar hidden, hovering the pointer at the window's leading edge now peeks the sidebar out **at the width it had before it was hidden**, and it slides back in when the pointer leaves the sidebar (or the window). The toggle shortcut, the toolbar button, and dragging the collapsed column out all stay in sync — one press always does what the screen suggests it should. The peek is toggleable in **Settings → Appearance → Sidebar → "Peek sidebar when hidden"** (default on).

## Why

Hiding the sidebar left two native NSSplitView mechanisms armed, and both misbehave under Macterm's transparent titlebar:

- the **edge-hover reveal** (`revealsOnEdgeHoverInFullscreen`, default-on for sidebar items and — despite the name — active in windowed mode on Tahoe), and
- the windowed **"proactive peek"** (`_NSSplitViewCollapsedInteractionsView` trackers installed over every collapsed pane).

Hovering the left edge popped a *minimum-width* sidebar overlay that crammed the traffic lights against the toolbar, never retracted on mouse-out, and left `columnVisibility` desynced from `AppState.sidebarVisible`, so re-hiding took two shortcut presses. The proactive peek also races any concurrent uncollapse of the same column: its target state is freed mid-engage and `mouseExited:` → `-[NSSplitView _cancelProactivePeek]` crashes on it (`EXC_BAD_ACCESS`, from a real crash log).

## How

- `WindowAppearance` disables both native mechanisms: the reveal by clearing the item flag and re-running `_updateHasItemToRevealOnEdgeHover` (which derives "anything to reveal" solely from that flag and `canCollapse*` — verified by disassembly), the proactive peek by replacing its single gate `_canDoSidebarProactivePeek` with a constant NO (no per-item opt-out exists, and its trackers reinstall on every collapse, so anything less re-races). Every SPI touch is probed with `responds(to:)`; an OS that removes them degrades to the old quirk, never a crash.
- `MainWindow` implements the peek itself with a small `onContinuousHover` state machine: a 12px leading-edge strip slides the column out through the exact `columnVisibility` path the shortcut uses — so SwiftUI restores its remembered width and the titlebar lays out as for a pinned sidebar — and the pointer moving past the sidebar's width (+8px grace) or leaving the window slides it back. Peeking never flips the user's toggle state; pressing the shortcut while peeked pins the sidebar, and hiding it under the pointer won't instantly re-peek (the pointer must leave the strip once).
- `columnVisibility` changes that bypass AppState (toolbar button, drag-out of the collapsed column) now mirror back into `appState.sidebarVisible`, ending the two-presses-to-hide desync.
- The `firstSplitView`/`owningSplitViewController` view-walk helpers move from `SettingsView` to `WindowAppearance` so both windows share one copy.


jcuhnoio:
Here's a demo:

https://github.com/user-attachments/assets/42f0a5fb-f28e-4741-8c9c-4916ea533e1c

thdxg:
@jcuhnoio can you add a toggle to enable/disable this in the settings?

jcuhnoio:
Added in 662498d — **Settings → Appearance → Sidebar → "Peek sidebar when hidden"** (default on), with the caption "Slides the hidden sidebar out while the pointer rests at the window's left edge." The gate sits in the hover handler rather than on the modifier, so flipping it off mid-peek still retracts the column.

image

thdxg:
@jcuhnoio I ran this locally and ran into a bug where the sidebar width shrinks to minimum width when peeked. This bug only occurs sometimes. Not sure what triggers this

https://github.com/user-attachments/assets/d29fada7-478e-4aec-ade0-ce7e6faecec6

jcuhnoio:
Two behavior refinements to the peek since the last review:

**1. The sidebar button pins a peeked sidebar instead of hiding it** (afda2ef). While a peek is out, the sidebar's *configuration* is still "hidden" — the peek is just a transient view of it. When someone reaches for the toolbar button in that moment, they're acting on the state they've set and remember, not on the frame currently on screen: their mental model says "the sidebar is hidden, show it". So the button's collapse-while-peeking is reinterpreted as "show", pinning the sidebar, rather than yanking it away from under the pointer. (The keyboard shortcut already behaved this way.)

**2. The peek's expand/collapse animations are never interrupted** (d330156). Reversing the column animation mid-flight turned out to be the root of two bugs: it corrupts `NavigationSplitView`'s stored width metric, so a dragged sidebar width silently reverts to the default after a few fast peek cycles — and a re-peek racing our own collapse could deliver the collapse's `onChange` while `isPeeking` was already true again, which is indistinguishable from the toolbar-button signature from #1 and wrongly pinned the sidebar with hover left dead. Both directions now settle (0.4s bound) before the opposite transition runs: a too-quick exit queues the retraction behind the expand, a too-quick re-entry queues the peek behind the collapse (re-checked against the pointer's last position, since a parked pointer generates no further hover events to retry on), and a shortcut hide stamps the same collapse cooldown. Fast thrashing now yields a clean out-in-out rhythm at animation speed instead of mid-flight reversals. This also supersedes 3722925 (feeding the remembered width back as `ideal`) — with no interrupted animations the width memory never corrupts, so that layer was solving the wrong problem and is reverted.

thdxg:
LGTM!

PR #230, opened 2026-08-07
Refine tab-drag UX: workspace-only merging, one-icon split rows, fading titles

thdxg:
Stacked on #228 — the base branch `feat/tab-drag-to-split` mirrors that PR's head (68d4990) so this diff shows only the refinements. Once #228 lands, retarget this PR to `main` and delete the mirror branch.

Refinements to the tab-drag/split UX, verified in the running app with synthesized drags and screenshots against a hermetic instance.

## Tab merging happens on the workspace surface only

Dropping a tab onto another tab's sidebar row merged them, with a hover preview that tried to mirror the destination's layout — but a sidebar row is a poor canvas for previewing a split, and the interaction kept fighting the sidebar's other drop roles (reorder between rows, append on headers). The workspace drop already does this job with an honest placement grammar (whole-edge / divider / local pane split, previewed at real size), so within-sidebar merging is removed: the sidebar keeps reordering and cross-project moves, and merging happens by dragging a tab's row into the workspace. The row drop target, its preview machinery, the now-unreachable `AppState.mergeTab(intoTab:side:)` / `TerminalTab.adoptTree(side:)`, and their tests go with it; the workspace merge path is untouched.

## No split overlay for a tab dragged over its own workspace

Dragging a tab's sidebar row over the workspace that is already rendering that tab showed the split drop preview, but releasing could only self-merge — which `AppState.mergeTab` refuses — so the overlay promised a split that never happens. `PaneDropContext` now carries the rendered tab's id and the leaf delegates reject a matching `MovableTab` payload outright: no preview, drag image animates back.

## Split rows: one tab icon, chromeless divided title segments

A split tab's row used to give each pane segment its own icon (agent logo or tab icon) while a renamed tab showed a single one — the same row read differently depending on how it was named. The segments are now a *title* variant inside the row's normal `Label`: every tab row carries exactly one icon (so the status-indicator/agent-icon logic applies to split rows too), and the per-pane titles share the row in equal widths, divided by hairlines, with no pill chrome.

## Every sidebar title fades on overflow instead of ellipsizing

`FadingText` (its own file, a drop-in one-line `Text` stand-in) fades an overflowing title at its trailing edge — an ellipsis burns several characters of an already-narrow slot. The layout element is a normal truncating text rendered invisibly so containers stay compressible (equal-width segments keep re-balancing as panes come and go); the visible copy draws in a layout-neutral overlay masked to the slot, and the ramp engages only when the text genuinely overflows (measured via `onGeometryChange`). Applied to tab titles, split segments, and project names alike. A single `rowTrailingInset` rule at each row's root keeps row content inside the selection highlight, whose region ends before the row content's edge.

## No idle-status indent on icon-less titles

With the tab icon set to None and the status indicator on, an idle tab's title sat nudged right: the "none" sentinel rendered as an invalid SF Symbol — invisible, but still reserving the Label's icon column. The label now gets an icon only while the status glyph actually draws something.

## Testing

- `mise run format` / `lint` clean; full unit suite passes (the removed feature's tests removed with it).
- Live-verified in a hermetic app instance with synthesized drags and screenshots: a tab dropped on another tab's row no longer merges (reorder/move still work); dragging a tab's row into the workspace still merges at the previewed placement; self-tab drag shows no overlay; split rows show one icon with divided segments; long tab titles, split-segment titles, and project names all fade inside the selection highlight; segments re-balance when a tab grows from 2 to 3 panes.


jcuhnoio:
These are all great catches!

thdxg:
Superseded by the PR above after #228's merge wedged this one's base in GitHub's stack machinery — same branch, retargeted to main.

PR #258, opened 2026-08-13
Publish a sortable image tag so the site can be deployed by Flux

thdxg:
## What

The website workflow publishes a third tag alongside `:latest` and `:sha-<short>`: a zero-padded UTC timestamp followed by the short sha, e.g. `20260813-142259-62701be`.

## Why

The site is being deployed to the k3s cluster at [macterm.thdxg.dev](https://macterm.thdxg.dev), with Flux's image automation keeping it current. Flux's `ImagePolicy` selects the newest image by **sorting tag strings**, and neither existing tag can carry that:

- `:latest` is one string forever — it never sorts to anything new.
- `:sha-<short>` sorts by whatever the abbreviated hash happens to be, which has no relation to build order.

A zero-padded UTC timestamp is the one format whose lexical order *is* its chronological order. This is the same tag the portfolio site already publishes, and the cluster consumes both the same way.

The two existing tags keep their jobs — `:latest` for a human `docker run`, `:sha-<short>` as the way back to a specific build after a bad one.

## How

One line added to the `publish` job's `docker/metadata-action` tags, matching the portfolio's:

```
type=raw,value={{date 'YYYYMMDD-HHmmss' tz='UTC'}}-{{sha}},enable={{is_default_branch}}
```

`enable={{is_default_branch}}` keeps it off PR builds, which push nothing anyway.

The format is a **contract with the homelab repo** — `apps/macterm/manifests/image.yaml` hard-codes `^\d{8}-\d{6}-[a-f0-9]+` as its `filterTags` pattern, so changing the format here without changing it there strands the deployment on its pinned tag. Both the workflow comment and `website/README.md` say so.

## Verified

- Tag template confirmed against the portfolio site's published tags (`20260812-142259-5b2c39d` and predecessors), which use the identical `metadata-action` expression — so `{{date}}`/`{{sha}}` are known to render this way rather than assumed.
- Confirmed `ghcr.io/thdxg/macterm/website` currently publishes only `latest` and four `sha-*` tags, i.e. nothing the policy pattern can match today. The first push to `main` after this merges produces the first selectable tag.
- Not run: `mise run format` / `lint` / `test`. This branch was prepared on Linux and they need macOS — the change is a workflow and a README, and CI on this PR exercises the workflow itself.

## Notes for reviewers

The companion deployment is a separate PR against `thdxg/homelab` (`apps/macterm/`). Merge order doesn't matter, but the site only starts auto-updating once **this** one has merged and produced one timestamped build. Until then the cluster runs the tag pinned in that PR (`:sha-62701be`, the current head of `main`) and the `ImagePolicy` reports no matching tag.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CfEJ5yKpFAcVAWuCbhEhcb)_

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -139,11 +139,19 @@ jobs:
         uses: docker/metadata-action@v6
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-          # :latest is what a deployment tracks; :sha-<short> is what it pins
-          # to, and the only way back to a specific build after a bad one.
+          # Three tags, each with a job. :latest is what a human pulls;
+          # :sha-<short> is the way back to a specific build after a bad one;
+          # and the UTC timestamp tag is what the cluster deploys, because
+          # Flux's ImagePolicy picks the newest image by sorting tag strings
+          # (`alphabetical: asc`) and neither of the other two orders. Keep the
+          # timestamp first and zero-padded — that is what makes lexical order
+          # and chronological order the same thing. Changing this format means
+          # changing the `filterTags` pattern in the homelab repo's
+          # apps/macterm/manifests/image.yaml to match.
           tags: |
             type=raw,value=latest,enable={{is_default_branch}}
             type=sha
+            type=raw,value={{date 'YYYYMMDD-HHmmss' tz='UTC'}}-{{sha}},enable={{is_default_branch}}
 
       - name: Create and push the manifest list
         working-directory: /tmp/digests
--- a/website/README.md
+++ b/website/README.md
@@ -100,8 +100,16 @@ and publishes one multi-arch manifest to
 `ghcr.io/thdxg/macterm/website`. Pull requests build both architectures without
 pushing, so a broken Dockerfile fails the PR rather than `:latest`.
 
-Two tags are published: `:latest` (what a deployment tracks) and
-`:sha-<short>` (what it pins to, and the only way back after a bad build).
+Three tags are published: `:latest` (what a human pulls), `:sha-<short>` (the
+way back to a specific build after a bad one), and a UTC timestamp tag,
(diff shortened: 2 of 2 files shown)


thdxg:
The red **Workers Builds: macterm** check is not from this PR, and I can't fix it from the repo.

#257 deleted `website/wrangler.toml` and `website/src/index.js` when it moved the site to a container image, but the **Cloudflare Workers Git integration is still connected to this repository**. It triggers a build on every push and PR, finds no Worker to build, and fails. It failed on `9e6a087` at 17:07:59Z — 1 second before the first GitHub Actions job even started, and against a diff that touches only a workflow file and a README. Every push to `main` will fail the same way until the integration is disconnected in the Cloudflare dashboard (Workers → `macterm` → Settings → Build → disconnect the repo, and delete the Worker if nothing else uses it). The check output carries no logs, only a dashboard link.

Every GitHub Actions check passed: **Build amd64** and **Build arm64** both green, so the Dockerfile still builds on both architectures with the new tag in place. **Publish** is correctly skipped — PRs build without pushing — which does mean the tag expression itself is first exercised on the merge to `main`. It's the same expression the portfolio site has been publishing with for months, so the risk is low, but the first `main` build is where to confirm a `20260813-…-<sha>` tag actually appears on the package.

One thing to check before merging, unrelated to CI: if the old Worker still holds a **route or custom domain** for `macterm.thdxg.dev`, it will intercept requests ahead of the wildcard DNS record and the cluster will never see them. DNS currently resolves `macterm.thdxg.dev` to the same Cloudflare IPs as the apex, which is what the wildcard CNAME would give — but a Workers custom domain resolves to Cloudflare IPs too, so DNS alone can't distinguish the two from outside. Worth confirming in the dashboard alongside disconnecting the build integration.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #262, opened 2026-08-16
feat: drop a pane on the sidebar to separate it into a tab

RafaelJaime:
Dragging a pane's grab handle onto the sidebar now separates it into its own tab — dropped between two tab rows it lands at that slot, dropped on a project header it appends (the only path for a collapsed or empty project, whose tab ForEach renders no rows to target). Cross-project drops work the same way tab-row drags already do.

## Why the sidebar could not be targeted before

`PaneDragDrop.swift` recorded the column as structurally unreachable, which is why "separate a pane into its own tab" shipped as the Separate Current Pane command rather than a drop. The real obstacle is narrower than that: a plain `NSPasteboardItem`'s custom-type data never reaches SwiftUI's `dropDestination(for:)` **import** path, so no `Transferable` destination in the sidebar ever saw the session.

So the sidebar doesn't try to receive the `NSDraggingSession` — it imports the pane's raw UUID bytes as one case of a `Transferable` union, the same import-only shape `SidebarDropItem` already used for tab and project payloads. The drag itself is unchanged: still an AppKit session writing the UUID eagerly, so the workspace leaves keep reading it straight off the drag pasteboard.

## Notes for review

- **Two unions, not a stacked destination.** The project header widens `SidebarDropItem` with a `.pane` case. The tab-list ForEach gets its own narrower `TabSlotDropItem` (tab + pane) so it can report the insertion offset; including the project payload there would put an insertion line inside a section a project can't land in. Each view still carries exactly one destination — the rule `SidebarDropItem`'s comment already documents.
- **Tab rows still carry no drop target**, so the native insertion line and tab reordering are untouched.
- **No model changes.** `AppState.separatePane` already accepted the destination index a sidebar drop reports, and its doc comment already described this caller. The `Pane` object is reused, so the surface and running shell survive.
- Dropping a tab's only pane is a no-op — it is already its own tab, and moving it across projects is what dragging its sidebar row does.
- `MovablePane(payload:)` is now the single decoder shared by the leaves' pasteboard read and the sidebar's import, covered by `PaneDragDropTests`.

## Verification

Built and exercised in the debug app on macOS 26.5 / Xcode 26.6: pane dropped between rows lands at that slot, on a header appends, shell keeps running, and tab reorder still shows its insertion line. `mise run format` and `mise run lint` are clean; the unit suite was left to CI.

### DIFF
--- a/Macterm/Views/PaneDragDrop.swift
+++ b/Macterm/Views/PaneDragDrop.swift
@@ -21,26 +21,45 @@ extension NSPasteboard.PasteboardType {
 
 /// The grab-handle drag as seen by its drop targets: the pane UUID
 /// `PaneDragSource` writes as 16 raw bytes under `.mactermPaneID`.
-/// Deliberately NOT `Transferable`: the drag is an AppKit `NSDraggingSession`
-/// whose custom-type pasteboard data never reaches SwiftUI's `dropDestination`
-/// import path, so the workspace leaves read the drag pasteboard directly
-/// through a raw `DropDelegate`. (This drag also cannot target the sidebar
-/// at all — no public drop mechanism in that column ever receives the
-/// session, which is why "separate a pane into its own tab" ships as the
-/// Separate Current Pane command rather than a sidebar drop.)
+///
+/// Deliberately NOT `Transferable` itself: the drag lifts as an AppKit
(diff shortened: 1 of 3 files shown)


thdxg:
@RafaelJaime This is a great addition! Thanks for your work! The PR looks good

Issue #104, opened 2026-06-22
Remote projects over SSH backed by zmx


onliner10:
# Remote projects over SSH backed by zmx

Macterm has been working really well for local projects, and I’d love to see a similar workflow for remote machines. I’d like to propose first-class remote projects backed by SSH + zmx.

## Use case

I usually work on one remote development machine. Today I open Macterm, SSH into it, and keep my actual work running there. I’d like that remote machine to be a real Macterm project instead of “a local pane running `ssh`”.

Ideally:

- A project can be a **remote machine**
- A Macterm tab maps to a persistent zmx session on that remote host
- Creating a new tab creates a new remote zmx session
- Work keeps running remotely when Macterm quits
- On app restart, Macterm SSHes back in and reattaches each tab to its saved zmx session

## UX

The sidebar “New Project” button could become:

```text
New Project
├─ Local Folder
└─ Remote Machine
```

“Remote Machine” would ask for:

- Display name
- SSH host / user / port
- Optional identity file, otherwise use the SSH agent / `~/.ssh/config`
- Remote working directory

The project appears in the sidebar like a local project, with a small server/SSH indicator.

## Implementation idea

PR #103 already introduces the core mechanism locally: each terminal has a stable `sessionID`, and the shell is launched through `zmx attach macterm-<sessionID>` so it can detach/reattach across app restarts.

For remote projects, the same idea could run **on the remote host**:

```sh
ssh devbox 'zmx attach macterm-<tab-sessionID>'
```

So the user-facing mapping would be:

```text
Macterm project  → remote machine / SSH target
Macterm tab      → one persistent zmx session on that machine
New tab          → creates a new remote zmx session
Close tab        → kills or detaches that remote zmx session, depending on the chosen lifecycle
```

zmx itself does not provide tabs, which seems like a good fit here: Macterm remains the source of truth for tabs and project layout, while zmx keeps each remote terminal session alive.

On restart, Macterm restores the workspace snapshot, SSHes into the remote machine, and reattaches each tab to its saved zmx session ID.

## Open questions

- Should Macterm require zmx to already exist on the remote host, or offer to upload/install a bundled Linux zmx binary?
- Should remote projects use SSH ControlMaster for faster splits/reconnects?
- How should connection failures be surfaced — pane banner, project-row status, or both?

I’d be happy to help work on this if it sounds aligned with the project direction.

thdxg:
[...]

PR #274, opened 2026-08-17
Serve the shell-integration ssh features natively, dropping the Ghostty.app CLI dependency

thdxg:
## What

Removes the last dependency on an installed Ghostty.app: the shell-integration `ssh` wrapper's `"$GHOSTTY_BIN_DIR/ghostty" +ssh` is now served by Macterm itself.

- **New offline CLI verb `macterm ssh`** (`CLI/SSHCommand.swift`) — ghostty's `+ssh` reimplemented natively: destination via `ssh -G`, first-connect `xterm-ghostty` terminfo install (`infocmp -x` from **our** bundled DB piped to a remote `tic -x -` over a throwaway ControlMaster), an install cache, then `execvp` of the real ssh with `SetEnv=TERM=…` + `SendEnv` forwarding for `COLORTERM`/`TERM_PROGRAM`/`TERM_PROGRAM_VERSION`. Flags mirror `ghostty +ssh` (`--terminfo`, `--forward-env`, `--cache`, `--ssh`, `--verbose`). It touches no control socket — works with no app running.
- **A `ghostty` shim** (`scripts/ghostty-shim.sh`) bundled at `Contents/Resources/ssh-bridge/ghostty` relays `+ssh` to `macterm ssh` and rejects everything else. `MactermConfig.regenerate` points `GHOSTTY_BIN_DIR` there.
- **Deleted**: `GhosttyCLI` (Ghostty.app detection + the blocking `ghostty +help` probe), its tests, and the Settings "Ghostty.app isn't installed" banner. `ssh-env`/`ssh-terminfo` are no longer force-disabled — they now just work; only `no-path` stays forced.
- Pure logic lives in `Macterm/System/SSHWrapper.swift`, compiled into both the app and CLI targets (the `ControlProtocol` pattern) and unit-tested through the app module. `RemoteTerminfo.entryName` delegates to it so the user-typed and Macterm-owned ssh paths can't promise different terminfo entries.

## Why

`ssh-env`/`ssh-terminfo` only worked when a new-enough Ghostty.app happened to be installed, and users who enabled them in their ghostty config silently got nothing. The remote-projects work (#104) already reimplemented the terminfo install natively (`RemoteTerminfo`); this applies the same reasoning to the ssh the user types into a local pane.

## Reviewer notes

- **Shim placement is load-bearing.** Verified in the fork's `Exec.zig`: libghostty *unconditionally* appends the host exe dir to pane PATHs, and `EnvironmentSetup` prepends `Resources/bin` — so the shim gets its own non-PATH dir (`ssh-bridge/`), and `no-path` stays forced permanently. A `ghostty` that only answers `+ssh` must never be reachable by name.
- **Cache key is a content hash** of the infocmp source, not ghostty's compile-time version (we have none — the entry arrives with each pinned GhosttyKit release). A GhosttyKit bump that changes the entry reinstalls once per host. One deliberate divergence from upstream: the cache is written right after a successful install rather than after the interactive ssh exits 0 — the install succeeding is the fact being cached, and this is what lets the wrapper end in `execvp` with no lingering process.
- **Behavior change**: the `path` feature (which put a real Ghostty.app CLI on pane PATHs when one was installed) is gone permanently.
- If the shim is ever missing from the bundle (broken build), `regenerate` falls back to forcing `no-ssh-env,no-ssh-terminfo` so the wrapper degrades to plain `ssh` instead of dying on a bad exec.

## Verification

- 16 new `SSHWrapperTests` (destination parsing mirrors ghostty's own ssh.zig tests; argv shapes; cache round-trips; the `entryName` cross-pin); full unit suite, lint, format, and the e2e suite (13 passed) all green.
- Live-drove the built CLI against an injected fake ssh (`--ssh=` is the seam, as upstream): dest resolution, install argv, cache write/hit, and byte-identical final argv to `ghostty +ssh` for every flag combination; shim relay + rejection verified.
- Hermetic app launch: generated overrides read `env = GHOSTTY_BIN_DIR=…/Resources/ssh-bridge` and `shell-integration-features = sudo,ssh-env,ssh-terminfo,no-title,no-path` — the user's own flags survive, only `no-path` forced.


thdxg:
**Real end-to-end verification added** (beyond the fake-ssh injection tests): ran the wrapper against a throwaway loopback `sshd` (own host key + authorized_keys, nothing touching `~/.ssh`) — and it caught a real bug that upstream `ghostty +ssh` also has: the install script was sent **bare**, and sshd hands remote commands to the *login shell*, so a nushell login shell rejected `2>&1` and every install failed. Fixed by applying the `RemoteSpawn` wire rule (`sh -c '<single-line, single-quote-free script>'`), with a regression test (`install_script_survives_non_posix_login_shells`).

With the fix, the full flow passes on the real OpenSSH stack:
- first connect through a **nushell login shell**: `tic` compiled the entry into `~/.terminfo/78/xterm-ghostty`, cache written
- reconnect: cache hit, no second install
- `TERM=xterm-ghostty` arrives via the pty request and `infocmp xterm-ghostty` resolves the freshly installed entry remotely
- `COLORTERM=truecolor` forwarded via `SendEnv` (server had `AcceptEnv`)
- exit-code propagation via the final `execvp`

Also merged `main` (kept the new `FullDiskAccessBanner`; the conflicting `GhosttyCLIBanner` + its `wantsCLISSHFeatures` gate are deleted by this PR — the features no longer depend on Ghostty.app, so there is nothing to warn about).

PR #282, opened 2026-08-20
fix: support current and legacy Ghostty config paths

halms:
## What

Use `~/.config/ghostty/config.ghostty` as the default Ghostty config path while retaining compatibility with the legacy `config` filename.

## Why

Ghostty renamed its XDG config file in 1.2.3. Macterm still defaulted to the legacy name, so users with a current Ghostty setup had to select the file manually.

## How

Choose `config.ghostty` when it exists or when neither candidate exists. Fall back to `config` only when it is the sole existing candidate. Explicitly saved custom paths remain unchanged. The Settings placeholder and documentation now use the current filename.

## Verified

## Notes for reviewers

The fallback matches Ghostty behavior without migrating or rewriting user files.


thdxg:
@halms Thanks for catching this. Ghostty also includes the `Application Support` directory for config on [macos](https://ghostty.org/docs/config#macos-specific-path-(macos-only):). Maybe we can iterate over possible directories including this and choose the first match?

halms:
Ah, yes, good point. I'll aim to recreate the exact ghostty logic:
- `Application Support` / `config.ghostty`
- `Application Support` / `config`
- `XDG_CONFIG_DIR` / `config.ghostty`
- `XDG_CONFIG_DIR` / `config`

halms:
This now turned into quite a bit of code for picking a config file, but yeah, we really should act predictibly correct here.

There is still a deviation though:
Ghostty actually loads all the files and merges them. I can look what it would take to do that, but it will be a bit more complex because Macterm expects a single file in a couple places. Will explore that option in another PR, then we can compare.

thdxg:
@halms Agreed. We can explore config merging in another pr. This pr looks good!

PR #283, opened 2026-08-20
fix: expose terminal surfaces to accessibility clients

halms:
## What

Expose terminal surfaces to macOS accessibility clients as editable text areas.

## Why

Dictation applications need the focused terminal to identify itself as an editable text target. At least Spokenly does not paste dictated text automatically unless the terminal view exposes this accessibility role.

## How

Make `GhosttyTerminalNSView` an accessibility element with the `.textArea` role, matching Ghostty.app. A focused unit test pins both accessibility properties.

## Verified

## Notes for reviewers

This adds the minimum role information needed by dictation clients. It does not attempt to expose terminal buffer contents through the accessibility value API.

### DIFF
--- a/Macterm/Views/Terminal/GhosttyTerminalNSView.swift
+++ b/Macterm/Views/Terminal/GhosttyTerminalNSView.swift
@@ -1,4 +1,5 @@
 import AppKit
+import CoreText
 import GhosttyKit
 import QuartzCore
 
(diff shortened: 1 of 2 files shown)


thdxg:
@halms This is a good direction, thanks. But Ghostty actually implements an accessibility client more comprehensively: https://github.com/ghostty-org/ghostty/blob/1ffa77c90bc47e8200e2c2c88b34ef6745ffd834/macos/Sources/Ghostty/Surface%20View/SurfaceView_AppKit.swift#L2320-L2420. 
Could you mirror Ghostty's implementation?

halms:
I think it should match now.

thdxg:
@halms Looks good! Thanks for your work

PR #284, opened 2026-08-20
fix: disable Sparkle in local builds

halms:
## What

Disable Sparkle in local builds that contain the placeholder EdDSA public key.

## Why

Local Release builds currently start Sparkle even though they cannot verify updates. This produces an "Unable to Check for Updates" alert. Debug and benchmark builds already avoid that failure.

## How

Start Sparkle only for non-Debug, non-benchmark builds with a nonempty, non-placeholder public key. CI release builds inject the real key and retain normal update behavior.

## Verified

## Notes for reviewers

The existing placeholder becomes an explicit local-build disable switch. Distributed release behavior is unchanged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -182,6 +182,10 @@ jobs:
             echo "error: tag '$GITHUB_REF_NAME' is not $EXPECTED; refusing to build a placeholder release" >&2
             exit 1
           fi
+          if [[ -z "$SPARKLE_ED_PUBLIC_KEY" || "$SPARKLE_ED_PUBLIC_KEY" == "SPARKLE_ED_PUBLIC_KEY_PLACEHOLDER" ]]; then
+            echo "error: SPARKLE_ED_PUBLIC_KEY is missing or still set to the local-build placeholder; refusing to build a release without auto-updates" >&2
+            exit 1
+          fi
           # Release builds invoke the build script directly to skip the
           # format/lint/test deps on `mise run build` — main is already
           # format/lint-gated by checks.yml and test-gated by test.yml before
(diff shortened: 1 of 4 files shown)


thdxg:
@halms Thanks for your work! Can you address these items:
- A missing `SPARKLE_ED_PUBLIC_KEY` secret becomes a silent, permanent
failure. `scripts/build.sh:19` falls back to the placeholder when the env var
is unset, and the release workflow never validates it — unlike the signing
certificate, which fails hard at `security find-identity` for exactly this
reason. Today that misconfiguration announces itself on every launch of the
shipped app; after this change it ships an app whose updater never starts, with
no alert and no log line. DMG signing and appcast publishing use the separate
`SPARKLE_ED_PRIVATE_KEY`, so both still succeed and the release looks clean.
Everyone who installs it is off the auto-update path until they manually
reinstall.
Since the placeholder is now load-bearing, could you add a guard alongside the
existing tag-format check in `release.yml`'s "Build universal release artifact"
step (or in `build.sh` when a release version is set) that refuses to build when
the key is empty or equals the placeholder? That converts the silent case back
into a loud one, in the right place this time.

- Pin the placeholder literal. `UpdaterAvailability.placeholderPublicKey` is
now a contract with `project.yml:96` and `scripts/build.sh:19`, and nothing ties
the three copies together — rename it in the two non-Swift copies and every local
Release build gets the launch-blocking alert back with no test failing. The
`betaUpdateChannel` test in this same file already sets the precedent (it reads
the literal out of `publish-appcast.sh`); asserting the Swift constant against
`project.yml` / `build.sh` would close it. The new test currently only compares
`placeholderPublicKey` to itself.

halms:
Updated, should be good now.

thdxg:
LGTM!

PR #288, opened 2026-08-20
feat: reorganize terminals between tabs by dragging in the sidebar

RafaelJaime:
Dragging a pane out to the sidebar separates it into its own tab (#262), but nothing put it back, and a split tab could only be moved whole. This makes the sidebar a place to reorganize terminals rather than only to reorder tabs.

## What changes

- **A pane or a tab dropped on a row joins that tab.** The row shows a vertical insertion line capped with a dot at the seam nearest the pointer, so the drop also picks the ORDER inside the destination — before, between, or after the terminals already there.
- **A split row's segments each drag their own pane.** A split tab reads as several terminals sharing a row; dragging the one you pointed at now moves that one. Dragging the row still moves the whole tab, from the icon and the space around the segments.
- **A row shows its panes' names whenever they fit.** The rule was a fixed "2 or 3 panes"; it is now measured against the width the row actually has, so a wide sidebar shows four names and a narrow one falls back to the pane count.

> **Screenshots to add:** the insertion line while joining, and a single segment being dragged.

## Notes for reviewers

- **The join band is a child of the row's title, not an overlay on the row.** Two earlier attempts placed it by measuring frames, and both leaked the tab's ICON back in as a droppable position — one because the measurement landed on `SidebarProjectRow`'s title (same shape, different row), the other because `.position` returns a view that fills its parent, so attaching `.onDrop` after it handed the whole row back as the target. Parenting removes the arithmetic and the ambiguity.
- **The band must stay off the row's top and bottom edges.** A target covering the whole row claims every drag before the tab ForEach's insertion mechanism sees it, which kills the native insertion line and tab reordering with it — the failure the row's own comment already recorded. Leaving the outer quarters uncovered is what lets one row mean three things: insert above, join, insert below.
- **It is a raw `DropDelegate`** because `dropDestination(for:)` reports no pointer location, and here the location IS the choice. `dropUpdated` also fires after `performDrop`, so it refuses to update unless a drag is still in flight; without that the indicator stuck to the row after every drop.
- **Seams come from the segments' drawn frames**, and an interior seam is the midpoint of the gap — exactly where the divider is. Even fractions of the row width land a few points beside it, which is visible when the whole point is to line up with the `|`.
- **`MovablePane` grows an export side.** It was import-only because the one drag writing it was an AppKit `NSDraggingSession` from the pane grab handle; a `.draggable` segment needs to write. Both write the same 16 bytes, so no drop target can tell them apart — but `.draggable` renders through an item provider rather than eagerly, so the workspace leaves get their async fallback back.
- **`AppState` gains the two calls the drop needs.** `mergeTab(intoTab:)` merges into a NAMED tab (the existing overload could only merge into the active one, which is never what a drop on a specific row means), and `mergePane` is its pane-sized sibling, routing a tab's only pane through the tab path so no emptied tab is left behind. Both reuse the `Pane` object, so surfaces and shells survive the move.

## Tests

`TabMergePlacementTests` covers the pure placement: where the seams are, which one a pointer picks (nearest, so both ends stay reachable), and that every seam resolves to a local split beside a named pane rather than a whole-edge drop — a root-edge target would make every seam mean the same thing. `AppStateTests` covers the model: a pane joining a named tab, a single-pane tab moving whole, a drop on its own tab as a no-op, and `mergeTab(intoTab:)` landing on the row that was dropped on rather than the active one.

## Verification

Built and exercised in the debug app on macOS 26.5 / Xcode 26.6, over several rounds of live testing: joining at each seam, dragging single segments, dropping a segment into the workspace, tab reordering still working from the row's edges, and the row's names appearing and collapsing as the sidebar is resized. `mise run format` and `mise run lint` are clean and the test target builds; **the unit suite itself has not been run locally**, so CI is the first place it runs.


thdxg:
> Dragging a pane out to the sidebar separates it into its own tab (#262), but nothing put it back, and a split tab could only be moved whole

You can drag a tab from the sidebar into the currently focused tab to create splits. I intentionally did not implement merging tabs within the sidebar because it creates inconsistent UX depending on the number of panes in that tab. I don't think there's much of a gap in tab/pane merging and splitting at this point in Macterm

PR #289, opened 2026-08-20
feat: show a project's uncommitted changes and branch without leaving the window

RafaelJaime:
A terminal shows what you typed, not what it changed. Reading a diff means leaving the window or spending a pane on `git diff`, and knowing which branch you are on means running `git branch` in whichever pane you were about to type in.

## What changes

- **A read-only changes panel docked beside the terminal**, opened from a toolbar button: the project's uncommitted files, each with its diff — line numbers, syntax coloring, and before/after thumbnails when the file is an image, for which git only says "Binary files differ".
- **The toolbar button carries the counts** (`+adds −dels`) so a clean tree is visible without opening anything.
- **The title bar names the focused terminal's branch**, beside the project path.

Nothing here writes to the repository: no staging, no commit, no discard.

> **Screenshots to add:** the panel open with a colored diff, an image file's before/after, and the title bar showing the branch.

## Notes for reviewers

- **The panel is a COLUMN in the detail area, not a pane in the split tree.** `SplitNode` is `.pane(Pane)` and a `Pane` owns a ghostty surface, so the tree cannot hold a non-terminal view without a much larger change. Being a sibling column rather than an overlay is also what lets the terminal reflow instead of being covered. If a non-terminal pane kind is on the roadmap, this is the piece that would move.
- **Loading is lazy per section, and that shape is the feature.** Fetching every diff up front spent whole cores on files nobody scrolled to (measured: 398% CPU on a large repository). Each section asks for its own diff when it scrolls into view, so opening a large repository costs one `git status`. Parsing runs off the main actor for the same reason — inline, it froze the panel on "loading" for as long as the parse took.
- **The tokenizer scans UTF-8 bytes rather than `Character`s.** A grapheme cluster comparison in that loop was enough to peg a core on an ordinary diff. Every boundary it recognizes is ASCII and any byte ≥ 0x80 counts as identifier content, so a multi-byte character can never be split across two tokens.
- **Shelling out to `/usr/bin/git` rather than linking libgit2**, since the app already depends on the user's toolchain and that path follows `xcode-select`. Two details are load-bearing: stdout is drained *while* the child runs (a diff exceeds the 64KB pipe buffer, and a child blocked writing to a full pipe never exits), and `GIT_TERMINAL_PROMPT=0` keeps a credential helper from parking a read-only call on a prompt.
- **The branch is read for the FOCUSED pane's working directory**, from the kernel via `ProcessInspector`, not from shell integration — so it is right for a shell that reports no cwd and for a full-screen program holding the foreground. A pane that has `cd`ed into a submodule or sibling worktree reports that branch, which is the question the title is answering.
- **Remote projects are excluded outright** — git runs on this Mac, and a remote project's path names a directory on another host. The toolbar button hides and the palette command disappears rather than opening something that could only explain itself.
- **A detached HEAD shows nothing rather than a branch called `HEAD`**, which is what `rev-parse --abbrev-ref` literally answers.

## Known rough edges

- The panel and the toolbar button read git independently, with separate caches and refresh cadences, so they can disagree for a moment after a change. Unifying them behind one reader is the obvious follow-up; they were built in that order and kept separable on purpose.
- Refreshes are coalesced off terminal activity (first event arms a read 1.5s later, with a floor between reads) rather than watching the filesystem.

## Tests

`GitStatusTests` covers the parsing that is easy to get subtly wrong: NUL framing (git quotes any path with a space in the human form), renames spending an extra field, both status columns kept separately so a staged-and-dirty file isn't misreported, and `+++`/`---` not being read as changed lines. `SyntaxHighlightTests` pins that joining the tokens reproduces the line exactly — a byte scanner that miscounts a boundary emits invalid UTF-8 — plus escaped quotes, unterminated strings (routine in a diff), and multi-byte identifiers. `GitBranchNamesTests` covers the branch string, including the detached-HEAD case.

## Verification

Built and exercised in the debug app on macOS 26.5 / Xcode 26.6 against real repositories, including a large one where the eager version was pathological. `mise run format` and `mise run lint` are clean and the test target builds; **the unit suite itself has not been run locally**, so CI is the first place it runs.


thdxg:
@RafaelJaime I appreciate your contribution! However, a git panel doesn't align with the direction I pursue with Macterm. I understand it is a nice-to-have feature to some users, and there are many other open source terminal emulators that do have features like a git panel, a browser, or more advanced AI harness integration.

I intentionally refrain from adding a feature that is simply an alternative to existing methods, and rather focus on things that can only be done at the terminal level. For example, showing git diff can be achieved without running extra commands by just having a dedicated tab running `lazygit` or equivalent programs in that project.

I am truly grateful for your interest and support to this project. And I'm sorry that I have to reject this one 🙏

PR #290, opened 2026-08-20
feat: let the tab spinner be hidden on agent tabs

### LINKED ISSUE #225: Ability to customize tab status indicator behavior
### Description

Currently, the behavior of the tab status indicator is such that:

- If enabled, shows a loader while the tab is busy, and shows a green dot when the process exits
- If disabled, shows neither

It would be useful to be able to have access to the following customization:
- Conditionally show the status indicator based on whether there's a running process AND some arbitrary user-defined condition
- Have granular control between showing the spinner and showing the green dot

### Motivation

The specific behavior I'm looking to achieve is the following:

When an agent process is running in a tab (claude code, codex, etc), don't show the spinner, because most agent CLIs have a built-in spinner in the tab title and it would be more useful to just see the agent icon (while for other processes the spinner is quite useful). However, when the process completes, still show the green dot because it's nice to see where you have "unread" messages from your agents.

I'm not an expert on Swift so I can't confidently verify this, but my guess is that the *general* case described in the issue description would be quite difficult to design & implement (e.g., where would the gating logic code live? is the behavior defined via the CLI or elsewhere? etc), while the specific case described above would be much simpler to add in the settings. However, I totally understand that it's a very niche request and it likely does not make sense to add a settings pane to cater to it. And truthfully, it's a minor QoL bump.

What do you think?

### Alternatives considered

I am not aware of any alternatives or workarounds.

RafaelJaime:
Closes #225.

An agent CLI draws its own spinner in the tab title, so ours repeats it — and it replaces the tab icon to do so, hiding the logo that says WHICH agent is working. Today the only cure is turning the indicator off, which also removes the finished dot, and that dot is how the requester spots agent tabs holding a reply they haven't read.

## What changes

A new **Hide spinner on agent tabs** toggle, under "Show tab status indicator" in Settings. With it on, a tab whose foreground is a recognized agent keeps showing the agent's logo while it works, instead of the spinner. Everything else is untouched: a plain command still spins, and the finished dot still appears everywhere.

Off by default — the indicator's normal behavior is what most users expect, and the issue itself scopes this as narrow.

> **Screenshots to add:** the new toggle in Settings, and a tab row with an agent running under each setting.

## Notes for reviewers

- **This is the narrowed version the issue's author proposed in his own follow-up comment**, not the general "indicator conditions" system the issue opens with. He wrote it off as hard to design (where would the condition live, how would it be expressed) and suggested exactly this toggle instead.
- **The spinner is suppressed; the dot is not.** That asymmetry is the whole request — an implementation hiding both would look correct while removing the signal he actually relies on. The tests pin it.
- **The rule moved out of the glyph's `body` into `TabStatusPolicy`**, a pure function over the three inputs (execution state, whether an agent holds the pane, and two preferences). Two of those are user settings, and that combination is what a view can't hold still for a test. That refactor is its own commit and changes no behavior.
- The toggle **dims when the indicator is off**, where it would have nothing to hide.

## Tests

`TabStatusPolicyTests` covers each combination: a plain running tab spins, an agent tab spins until the preference is on, the preference suppresses the spinner only on agent tabs, the finished dot survives it, and the indicator's own switch still wins over everything.

## Verification

Built and exercised in the debug app on macOS 26.5 / Xcode 26.6. `mise run format` and `mise run lint` are clean and the test target builds; **the unit suite itself has not been run locally**, so CI is the first place it runs.


thdxg:
@RafaelJaime I just merged a pr for the same issue... haha
I added a toggle "Show spinner over agent icons" in settings, which seems similar to what this pr tries to achieve.
Thanks for your work tho!

RafaelJaime:
> @RafaelJaime I just merged a pr for the same issue... haha I added a toggle "Show spinner over agent icons" in settings, which seems similar to what this pr tries to achieve. Thanks for your work tho!

Ouch, I will close it then!

PR #292, opened 2026-08-20
Match Ghostty's config file merging

halms:
## What

Loads automatic user configuration through libghostty's own default-file loader, then mirrors its root-file order for Macterm's raw-text consumers. Settings retains custom-file and disabled modes.

## Why

PR #282 finds the preferred edit target, but Ghostty actually merges its XDG and Application Support files. Selecting only one drops valid overrides.

## How

`GhosttyConfigSource` centralizes mode selection and raw inspection. Runtime loading and automatic edit-path selection delegate to the pinned GhosttyKit APIs, preserving warnings and first-run template creation. Tests and benchmark/e2e runs stay isolated from real user config.

## Verified

- [x] `mise run e2e --verbose` passes (13 tests)

## Notes for reviewers

Rebased onto current `main` after #282, #283, and #284 merged. The PR is a single config-merging commit.


thdxg:
Screenshot 2026-08-20 at 5 19 02 PM

@halms Can we maybe add a list of detected files under the toggle? Also, I think users should be able to customize config file locations because they might wanna use different configs for Ghostty and Macterm. So I think it would be great to have a list of config files that are populated with default locations initially, but also customizable.

halms:
I was thinking: if we go for a custom config, 1 should be enough, but actually makes sense.
Maybe users want to use Ghostty config as a base and then some custom Macterm changes.

I'd propose the following:
- toggle for default Ghostty files with list of detected files
- custom field to add zero or more custom files in addition
- drop the technical comment about XDG and Application support, just the list should be okay.

halms:
I've now done roughly 30 iterations until I found a UI I liked.

image

thdxg:
@halms I like this idea. But I think we should make the default config paths regular items in the list so users can remove or reorder. So when the first-time user launches the app, the list populates with default locations. And for existing users, they would just see the config path they've been using.

Screenshot 2026-08-20 at 10 10 47 PM

I've also tweaked the design to match other lists in the Projects tab in settings. What do you think about this? If you're okay with it, I can commit to this PR.

halms:
My question with this is: what does the `Load Ghostty config files` toggle do once the files are in there?
What if a user turns on, then deleted one file, then turns off again, then turns on again, ...

I'd rather treat this as one atomic decision:
"Do I want to use my Ghostty config, at least as a base"

The edge cases where a user just wants to use part of the Ghostty config should be covered well enough by the possibility to turn the toggle off and then add the paths manually (but I'd assume this is very rare)

I like the design of the additional files.

And yes, feel free to push here.

thdxg:
@halms I think the `Load Ghostty config files` message might be a little misleading. It should be more like `Use Ghostty config files` toggle. So when turned on, Macterm will use the listed config files. When turned off, Macterm will not read the config files.
I'll commit to this pr so you can test my idea!

thdxg:
Toggle demo:

https://github.com/user-attachments/assets/419d5a78-8c3d-4820-87d0-22c4b3fa4da0

halms:
Ah, the toggle now only does "use no files at all".
I don't know if we even need that. If one wants no config files to be used, they can just delete it.

I'd still prefer basically a toggle-able list to auto-discover Ghostty's default file location – that also works if a file changes location later, etc.
And I think it's fine to just list the discovered ones in small grey.
Screenshot 2026-08-21 at 20 51 23
This is the "do whatever Ghosttty is configured to do" toggle.

And then below it the "additional" or "custom" files with the UI you proposed.

But it's your call.

thdxg:
Yes now that I think about it, having a toggle is redundant with just not listing any config files. What if we just remove the toggle from my design? The default config files can still be auto-discovered on first launch

halms:
The simplest case I can think of (others are rarer) that breaks:
- User has `.config/ghostty/config`
- User uses Macterm, it discovers and store this file, uses same config as Ghostty
- User sees that the new default is `.config/ghostty/config.ghostty` and renames the file to the new format
- Ghostty discovers the new file and continues to work the same
- Macterm doesn't discover the new file, thus works without configuration

thdxg:
I see. What about this version?

https://github.com/user-attachments/assets/33788943-b6e5-4552-87a2-f7e3a2fba279

When using default locations, they will always be automatically detected just like Ghostty does

halms:
I think it's fine.

But if you ask me, the use case
> I want to use my Ghostty config (wherever that is) but change a small thing about it (e.g. font)

Is more straightforward with a "use Ghostty config" (toggle-able) plus any custom files vs. the choice _either_ Ghostty default _or_ custom files.

Maybe: use the layout from your last version, but make the toggle "Load Ghostty's default config files".
If it's on (probably the default), then the list below shows the discovered files first, without the controls at the right, but maybe something like "Ghostty default config file"

Sorry for being annoying 😅

thdxg:
@halms I like this design! I think this is now good to merge. Thanks so much for your contribution

PR #298, opened 2026-08-24
fix: play sounds for terminal notifications

halms:
Match Ghostty and standard macOS app behavior by playing the default system sound for notifications.

Not everyone might want notification sounds, but they can be turned off for Macterm in macOS System Settings → Notifications.


thdxg:
@halms Thanks! LGTM

PR #329, opened 2026-09-01
Add an option to attach all project terminals on launch

nabilfreeman:
Adds an opt-in setting to attach every restored project's terminals when Macterm opens.

zmx sessions already keep running while Macterm is closed. The problem this solves is on the client side: background projects normally stay detached until they are selected, so a dashboard setup requires clicking through every project once after launch.

When enabled, Macterm attaches the saved terminals using its existing staggered background attachment path. It does not expand projects, change the selected project or tab, or alter wake and network reconnect behavior.

The setting is off by default because it may open multiple local shells or SSH connections. If a remote connection needs interactive authentication, its tab still needs to be opened so the prompt is visible.

Verified with:

- mise run format --verbose
- mise run lint --verbose
- mise run test --verbose
- mise run e2e --verbose (32 passed, 1 Docker-only test skipped)


thdxg:
@nabilfreeman Thanks for your contribution! I want to clarify something here. When quit and relaunched, all unfocused projects are simply detached from zmx, meaning that the zmx sessions are still running in background whether the project is focused or not. You mentioned current scheme is awkward when several projects are meant to keep running together, but the projects do run no matter what happens in the UI. I'm not sure I understand what benefit you are trying to achieve here.

nabilfreeman:
Sorry, I think the scope crept a little, and I also explained the problem poorly. You are right that the zmx sessions and agents keep running. What I am trying to restore is MacTerm's connection to them.

After relaunch, and sometimes after sleep or a network reconnect, the background projects are not reattached in the UI. I have to click through each project once before its tabs reconnect. I use MacTerm as a dashboard for several long-running sessions, so I would like an opt-in way to have them ready without that manual pass.

I'll look over the code again and see if this can be simplified. I will also bring it up to date with your upstream changes.

thdxg:
@nabilfreeman Is there a benefit to attaching to panes that are not visible in the UI?

PR #330, opened 2026-09-01
Rename tabs by double-clicking their sidebar title

nabilfreeman:
## What

Double-clicking a tab name in the sidebar now opens the existing inline rename field.

## Why

Renaming currently sits behind the context menu. The tab name is already the obvious place to edit it.

## How

The title uses the same selection binding as the rest of the row, so its first click selects the tab immediately. A second click opens the existing rename field. Context menu and keyboard behaviour stay unchanged.

## Verified

The end-to-end suite passes: 15 tests passed and the Docker-only remote fixture was skipped.

## Notes for reviewers

The screenshot uses temporary local projects. “Design system” is in the inline editor after double-clicking its title.

Sidebar tab in edit mode after a double-click

### DIFF
--- a/Macterm/App/FocusRestoration.swift
+++ b/Macterm/App/FocusRestoration.swift
@@ -15,6 +15,21 @@ enum FocusRestoration {
     private static let maxAttempts = 40
     private static let retryInterval: TimeInterval = 0.05
 
+    /// Set while a sidebar row's inline rename field is open (see
+    /// `SidebarPresentationState.beginRename`).
+    ///
+    /// Restoration is a *retry loop*, so it can land long after the call that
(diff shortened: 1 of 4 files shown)


thdxg:
@nabilfreeman Thanks for this! I've made some fixes

nabilfreeman:
@thdxg that's great! Thank you for accepting my submission! I have opened some other ones too.

I hope you don't mind, I am running these fixes in a local copy that fits my own workflow, and I will open PRs for anything that is well scoped, not a hack and beneficial to other users.

Then you can decide if they are worth integrating.

thdxg:
@nabilfreeman Of course! I appreciate your contribution! I've left comments on your unmerged PRs

PR #331, opened 2026-09-01
Add a new-tab button to project rows

nabilfreeman:
## What

Hovering a project row in the sidebar shows a small plus button. Clicking it opens a new tab inside that project.

Appearance settings now includes an option to show or hide the button. It is enabled by default.

## Why

Creating a tab currently sits behind the context menu or keyboard shortcut. The hover button keeps the action within reach for mouse users, while the new setting lets people keep the original sidebar if they prefer it.

## How

The button only enters the row layout while that project is hovered. When idle, the ordinary project row is left unchanged, so the hidden control takes up no sidebar space.

MacTerm already fades titles that genuinely overflow their row. This change leaves that behavior untouched. The earlier implementation kept the invisible button in the layout, which shortened every project title by about 24 points and made the existing fade start early. This version only adds the button and its spacing during hover.

The hover button and the existing context-menu action use the same new-tab path.

## Verified

- [x] End-to-end suite: 25 passed, with the Docker-only remote fixture skipped
- [x] Confirmed the idle row keeps its full width
- [x] Clicking the hover button created and selected a new fourth tab
- [x] Turning the preference off removed the button from a hovered row immediately
- [x] The existing project context-menu action still works

## Notes for reviewers

Expanded project with the hover action visible:

Synthetic MacTerm workspace with the Atlas Web project hovered and its new-tab button visible

Idle row, with no space reserved for the button:

Selected project row at rest with no new-tab button or reserved spacing

The same row while hovered:

Selected project row on hover with the new-tab button visible

After clicking the button, a fourth tab is created and selected:

Expanded project after the hover button created and selected a fourth zsh tab

The new Appearance preference:

Appearance settings with Show new tab button on projects enabled

The existing project context menu still offers the same action:

Project context menu with New Tab as its first action

With the preference disabled, hovering leaves the row unchanged:

Hovered expanded project with the new-tab preference disabled and no button visible


thdxg:
@nabilfreeman Creating a new tab can also be done with `⌘t` keybind, which is more convenient than clicking imo. Not sure if this button is worth taking up a portion of sidebar even when idle?

Screenshot 2026-09-01 at 8 38 49 PM

nabilfreeman:
@thdxg see my comment here, this excludes users who cannot/choose not to use keyboard:

https://github.com/thdxg/macterm/pull/338#issuecomment-5526407670

nabilfreeman:
> @nabilfreeman Creating a new tab can also be done with `⌘t` keybind, which is more convenient than clicking imo. Not sure if this button is worth taking up a portion of sidebar even when idle?
> 
> Screenshot 2026-09-01 at 8 38 49 PM

I think it is possible to remove that fade behaviour and only add the fade when the project is moused over, if that is a good compromise?

thdxg:
@nabilfreeman Yes I think that would work. Also it would be great to have an option in settings to enable/disable the button

nabilfreeman:
That sounds good. I'll add an option in Settings to enable or disable the button, and make sure it only appears on hover without reserving sidebar space while hidden.

nabilfreeman:
Updated. The button now reserves no space while hidden, and Appearance settings includes an option to turn it off. I refreshed the screenshots and verified the hover button, preference, and existing context-menu action.

thdxg:
@nabilfreeman Thanks! I'll include this in the next release

PR #337, opened 2026-09-02
feat: add directory preferences for new tabs and splits

### LINKED ISSUE #334: Add a new tab directory preference
### Description

New tabs currently start in the project's configured directory.

Add a preference with two options:

- Project
- Active pane

Project remains the default. Active pane uses the focused pane's current working directory. If that directory is unavailable, or the target project has no active pane, New Tab falls back to the project directory.

The preference should apply to New Tab from the keyboard, menu, command palette, and project context menu.

I'm happy to implement this if the behavior fits the project.

### Motivation

When working below a project root, Cmd+T returns to the root every time. Split panes already inherit the focused pane's directory. This preference would offer the same behavior for tabs without changing the existing default.

### Alternatives considered

A separate "New Tab Here" action would provide a one-time override, but directory inheritance is usually a consistent user preference.

"Replace project path with current directory" affects future tabs, but it also changes the project's saved root.

Pavior0:
## What

Adds separate `New tab directory` and `New split directory` settings. Both offer `Project` and `Active pane`.

Tabs still default to `Project`. Splits still default to `Active pane`, so existing behavior does not change. All existing UI entry points and the `macterm tab new` and `macterm pane split` commands use these settings.

## Why

`Cmd+T` always returned to the project root, even when the active shell had moved into a nested directory. This makes active-pane inheritance configurable without changing the existing defaults for tabs or splits.

Closes #334

## How

- `AppState` resolves the new terminal directory before creating a tab or pane.
- The selected cwd does not change the project-scoped zmx session slug.
- CLI tab creation reads the target workspace's focused pane. CLI splits read the selected source pane and fall back to the focused pane when no selector is given.
- Remote panes fall back to the configured `host:path`.

## Verified

- [x] `mise run format --verbose`, `mise run lint --verbose`, and `mise run test --verbose` pass
- [x] Built and ran the Debug app and confirmed both settings in the UI
- [x] Added behavior tests for tabs, splits, zmx session identity, and remote fallback
- [x] `mise run e2e --verbose` passes locally. The opt-in Docker remote reconnect test was skipped

The bundled CLI was also exercised against an isolated Debug app. This covered nested tab cwd inheritance, explicit-source splits, Project-mode splits, and remote fallback.

## Notes for reviewers

Quick Terminal is intentionally unchanged. It has no project workspace, starts at `$HOME`, and uses the dedicated `quick` zmx slug. Its splits continue to inherit the source pane cwd.

Codex assisted with implementation and test execution under direct human direction.

### DIFF
--- a/Macterm/App/AppCommandActions.swift
+++ b/Macterm/App/AppCommandActions.swift
@@ -115,13 +115,25 @@ extension AppCommand {
             return { ctx.appState.separatePane(paneID, toProject: projectID, destPath: destPath, at: index) }
         case .splitRight:
             guard let projectID else { return nil }
-            return { ctx.appState.splitPane(direction: .horizontal, projectID: projectID) }
+            return {
+                ctx.appState.splitPane(
+                    direction: .horizontal,
+                    projectID: projectID,
+                    projects: ctx.projectStore.projects
+                )
+            }
         case .splitDown:
             guard let projectID else { return nil }
-            return { ctx.appState.splitPane(direction: .vertical, projectID: projectID) }
+            return {
+                ctx.appState.splitPane(
+                    direction: .vertical,
+                    projectID: projectID,
+                    projects: ctx.projectStore.projects
+                )
+            }
         case .splitAuto:
             guard let projectID else { return nil }
-            return { ctx.appState.autoSplitPane(projectID: projectID) }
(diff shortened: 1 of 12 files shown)


thdxg:
@Pavior0 Thanks for this! The overall design is good. I'll add some minor fixes and merge this!

PR #338, opened 2026-09-02
Add terminal scroll navigation to the context menu

nabilfreeman:
## What

Adds "Jump to Top" and "Jump to Bottom" to the terminal context menu.

## Why

Long output can leave the live edge a long way away. These actions keep the mouse-accessible path without adding custom terminal chrome.

## How

Both menu items use Ghostty's existing scroll actions. Each one is disabled when the terminal is already at that edge.

## Verified

- [x] Focused tests cover no scrollback, both boundaries, and the middle of the buffer
- [x] Built and ran the app with synthetic output
- [x] Used both menu actions and confirmed the viewport moved to the expected edge

## Notes for reviewers

This replaces the floating button from the earlier version of the PR and follows the context-menu approach suggested in the discussion. The existing scrollbar behavior is unchanged.

The screenshots use temporary local projects and synthetic terminal output.

Scrolled between the two edges, with both actions available:

Terminal scrolled into synthetic output with Jump to Top and Jump to Bottom in the context menu

Context menu detail:

Terminal context menu with both scroll actions available

At the top, only Jump to Bottom is available:

Terminal context menu at the top of scrollback with Jump to Top disabled

At the live edge, only Jump to Top is available:

Terminal context menu at the live edge with Jump to Bottom disabled


thdxg:
@nabilfreeman Ghostty already has this feature as a key-bindable action: https://ghostty.org/docs/config/keybind/reference#scroll_to_bottom. I think having a UI for this instead of a simple keybind in ghostty config is a bit unnecessary.

nabilfreeman:
@thdxg I actually don't use a keyboard much due to RSI, I do everything these days with a trackball and dictation. Probably explains why my PRs have been focused on adding missing accessibility UI!

nabilfreeman:
I also predict you are about to find a lot more MacTerm users who do not even know what Ghostty is

thdxg:
@nabilfreeman I see. I haven't thought from that perspective. But I think we need to make some compromise here. I try to minimize custom chromes in the UI as much as possible, since I prioritize the native experience a lot. Something like a New tab button in your other PR is subtle and blends well with the native sidebar. However, I think it's quite difficult to achieve the same level of native feeling with this button.
Would it suffice if we add "Jump to bottom" and "Jump to top" actions in the terminal context menu?

nabilfreeman:
Yes, that works well. Thanks for meeting me halfway. I'll replace the floating control with "Jump to Top" and "Jump to Bottom" in the terminal context menu. That keeps the mouse-accessible path without adding custom chrome.

nabilfreeman:
Updated. The floating control is gone, and the terminal context menu now has Jump to Top and Jump to Bottom. Each action is disabled when the terminal is already at that edge. I also refreshed the screenshots and verified both actions against live scrollback.

thdxg:
Pushed one commit on top (74bfb77) with the review follow-ups, so the description above should be read with these in mind:

- **Scroller dedupe resync.** `SurfaceScrollView` kept its own last-sent row, so after a core-driven jump a later wheel or scroller move landing on exactly that row was dropped. `applyScrollbar` now resyncs it from the snapshot. This was already reachable via ghostty keybinds; the menu is the first mouse-only path onto it.
- **`autoenablesItems = false` also fixes Paste.** With auto-enabling on and no `validateMenuItem` on the target, AppKit re-enabled every item, so `paste.isEnabled = hasPasteboardContent()` never took effect before this PR. Worth knowing when reading the diff; there's a comment on the line now.
- The availability predicates moved onto `ScrollbarSnapshot` (`hasScrollback`, `maxScrollableRow`, `canScrollUp`, `canScrollDown`) so the menu, wheel handling, scroller clamping and `pane inspect` share one definition; the tests assert the properties directly.
- The binding-action call is one `sendBindingAction` helper that logs when libghostty rejects the action; the four search helpers use it too.

`mise run format`, `lint` and `test` pass locally. The `benchmark:regression` label is runner noise (nothing here runs outside a right-click); the next run should clear it.

thdxg:
@nabilfreeman Thanks! I made some minor adjustments, and this is ready to merge!

PR #340, opened 2026-09-03
fix: split MactermApp's scene body so it type-checks

micha-f:
## What

Splits `MactermApp.body`'s alert chain into three `ViewModifier`s  (`CloseConfirmationAlerts`, `ProjectConfirmationAlerts`, `LayoutAlerts`) beside the existing `AppColorScheme`.

## Why

`MactermApp.swift` won't compile on Xcode 26.3: *"unable to type-check this expression in reasonable time"* at `var body: some Scene`. The chain is `MainWindow()` + 2 `.environment` + `.modifier` + 7 `.alert` + `.onAppear`, and each `.alert` carries 3 generic params and 2 ViewBuilder closures.

CI never sees it: `test.yml` pins no Xcode and `macos-26` defaults to 26.6, where the same expression is fine. CONTRIBUTING.md documents "a full install of Xcode 26" as the requirement, so 26.3 is a supported configuration. Incremental builds hide it too. Only a clean build, or an edit to the file itself, recompiles it.

## How

Verbatim move: same order, same bindings, no behavior change. The `DialogHost` gating comment travels with the alerts it explains.

The modifiers carry a three-line doc comment recording that the split is a constraint rather than tidying, so it doesn't get inlined back.

## Verified

Reproduced on clean `main` first. Stashed all local work, unmodified checkout, identical error at `MactermApp.swift:48`. `mise run test` 86/86, format + lint clean on Xcode 26.3. The seven alerts weren't fired by hand; nothing inside them changed.

Possible follow-up: pin Xcode in CI.


thdxg:
@micha-f Thanks for catching this! I made some doc comment adjustments.

PR #341, opened 2026-09-03
feat: tag projects with a color

micha-f:
## What

An optional color per project. It tints the project's sidebar icon and the icon on each of its tabs. Set from the project's context menu or Settings → Projects. Optional auto-assignment on creation.

image

## Why

With several projects open a color answers at a glance which project a tab belongs to.

## How

- The tag rides a glyph each row already draws — nothing is added to a row, no gutter reserved, so a tagged sidebar occupies exactly the space an untagged one does.
- A tagged project claims its agent logos too (`SidebarRowIcon.agentTint` over `AgentIcon.brandColor`): the logo's shape says which agent, the color says which project. Untagged projects keep their brand colors.
- No row ever carries a stripe: with sidebar icons set to `None` there is nothing to tint, so tags don't show.
- `Project.colorName` in `projects.json`: a String, not the enum, so an unknown value reads as untagged instead of failing the row's decode. An older build drops the key silently on its next save.
- Eight fixed system colors, not the ghostty palette. A theme puts any color in any ANSI slot (Rose Pine's "cyan" is a rosy beige).
- Menu swatches come from `.pickerStyle(.palette)`: a tinted SF Symbol renders in the menu's own tint, and a non-template `NSImage` is dropped by SwiftUI's menu bridge.
- Auto-assign (Settings → Appearance, default off): least-used wins, ties broken by case order, so the first eight projects get distinct colors and only the ninth repeats. Read at creation only; flipping it never retags existing projects.

## Verified

`mise run test` green (1255 tests), format + lint clean.

Exercised in the app: colors set from the context menu and from Settings; the project icon and its tab icons take the tint, agent logos included; with icons set to `None` nothing is tinted; auto-assign picked the least-used color; sidebar drag-and-drop still works.


thdxg:
@micha-f Hi, thanks for your work! I see how project colors can be useful to group them visually. Instead of the color strip for each tabs, can we color the tab icons?

Also, the folder icon in toolbar doesn't look good in macOS 27. I'm also not sure if having the icon there is useful.

Screenshot 2026-09-03 at 7 59 42 PM

micha-f:
Both done, thanks. Screenshots updated.

- **Tab icons carry the color** now. Nothing is added to a row, so tagged and untagged sidebars line up identically (the stripe had indented tagged rows further).
- A tagged project also claims its **agent logos**. Keeping Anthropic coral next to a red project's icons read as a mistake rather than as a brand, so: the logo's shape says which agent, the color says which project. Untagged projects are untouched.
- With **icons set to `None`** a tab row shows no tag rather than falling back to a stripe; the project row still stripes, so there is something to tint.
- **Titlebar is a plain dot** instead of the folder. The icon version handed an SF Symbol to a `ToolbarItem`, which lays its content out at control size (that's why it outsized the title beside it). I kept a simple indicator there because it answers "which project is this window on" without looking down at the sidebar, but say the word and it comes out.

thdxg:
When a project has no icon, I think it should not have anything preceding the project name. Otherwise, it would create inconsistent visuals between colored vs. not-colored projects like this:

Screenshot 2026-09-04 at 4 17 46 PM

Also, the toolbar icon still looks broken, and I don't think it serves much purpose.

Screenshot 2026-09-04 at 4 21 57 PM

micha-f:
Both dropped. The tag is only ever a tint on an icon the row already draws, the titlebar indicator is gone.

thdxg:
@micha-f Thanks! I made minor fixes. Looks good to merge now!

Issue #345, opened 2026-09-04
Consider multi-window support


Pavior0:
### Description

Macterm is intentionally single-window today. I keep running into two cases where that feels limiting: separate monitors and macOS Spaces.

A familiar browser pattern, such as Arc's, is to distinguish a disposable blank window from a window tied to existing workspace state. Macterm could leave room for the same distinction:

- `New Blank Window`: opens a temporary scratch window. Its tabs, panes, and project state are not saved. Closing it drops the state, and it is not restored on relaunch.
- `New Window`: keep this action available for a future synced window. For example, it could open the current tab or project in another window, either by moving it or by attaching another client to the same zmx session.

The synced version seems plausible because [zmx already supports multiple clients per session](https://github.com/neurosnap/zmx#features). Macterm would still own the window and tab UI.

This is just a product direction, not a complete spec. I am mainly curious whether this fits Macterm's model.

### Motivation

With one window, Macterm can effectively be used from only one macOS Space at a time. Multiple windows would let me keep one terminal context in Space 1 and another in Space 2, while still using the same app. It would also make multi-monitor setups much less awkward.

A blank window should be cheap and disposable. A synced window should be an explicit action with different persistence and session behavior.

There are some terminal-specific tradeoffs. zmx has one leader client controlling the PTY size, so two synced windows would not have completely independent dimensions. Input, selection, scroll position, and other view state would still belong to each window. That seems reasonable if the UI makes the distinction clear.

### Alternatives considered

- Keep using tabs and splits only. This keeps the current model simple, but does not solve the Spaces use case.
- Add `New Blank Window` first. This seems like the smallest useful step.
- Try synced tabs first, without syncing the whole project or sidebar state.
- Run a second Macterm process with `open -n`. That gives separate app state and does not provide the same multi-window experience.

thdxg:
@Pavior0 Thanks for the suggestion. I think having multiple windows is feasible. But I think it would be easier to just have synced windows (while each window owns the focus state) and not blank windows. Currently every pane corresponds to a persistent zmx session, but introducing the "blank window" would mean extra machinery where panes have to bypass zmx.

What do you think of having multiple synced window support?

Pavior0:
[...]

PR #350, opened 2026-09-06
Track which mirror drives a session, and dim the ones that don't

thdxg:
> Stacked on #349, which is stacked on #348. Review those first; the diff here is just this commit.

## What

`AppState` now tracks which pane is zmx's **leader** for each mirrored session, and non-leader mirrors are dimmed.

## Why

zmx allows several clients per session but only one leader, whose window size it applies to the pty — a non-leader's resize is dropped outright (`handleResize`). Every client still receives the same broadcast output, so a non-leader mirror is **live but laid out for another pane's geometry**. After #349 nothing in the UI said so, which is the confusing half of mirroring.

Refs #345.

## How

**Leadership is recorded where zmx's daemon would hand it over anyway**, so the model tracks the daemon rather than inventing a parallel truth:

- **on focus** — any real keystroke requires focus, and a keystroke is exactly what zmx switches leader on (`util.isUserInput`);
- **on `pane run` / `pane key`** — injected input flips the daemon's leader with no focus change to notice it by;
- **on mirror creation** — the source keeps it, because `handleInit` sets a leader only when there is none, so a second client attaching leaves the pty size where it was.

With no record (a restored pair that attached before we tracked anything) it infers the first pane in tree order — zmx makes the first client to attach the leader and restore warms in tree order, so they agree in practice, and any real keystroke resynchronises them. Documented as best-effort rather than authoritative.

**Takeover already works with no zmx change.** Typing into a dimmed mirror makes it the leader and the pty resizes to it — that is the daemon's own behaviour. Clicking to take over needs a claim message zmx has no tag for; that is the next PR.

**The dim reuses `MactermTheme.dimOverlay` but not the split-dim's gate**, which is wrong for a mirror on both counts: `isSplit` is false for a single-pane tab *and* for a zoomed pane — exactly the common mirrored case — and the `adaptiveBackgroundColor == nil` exemption must not apply, since the point is to say "not the live size" whatever colour the TUI painted.

**A non-leader mirror is excluded from adaptive background inference.** Its painted region describes the *leader's* window, so inferring from it would retint this window to whatever the other one is showing. It gets the same freeze-on-last-confirmed treatment as a viewer overlay, but through a **separate parameter**: both make a frame untrustworthy, for reasons that must not be conflated — nothing is drawn over the screen model here, the model itself is foreign. The flag reaches the NSView via `TerminalPane.configure`, the existing seam for pane-aware state (the NSView has no back-reference to its `Pane`).

## Two session-identity bugs fixed along the way

Both needed a leader to point at, which is why they land here rather than earlier:

- **`session list` collapsed a mirrored session to one arbitrary pane id** — `map[sessionName] = paneID`, last writer winning in Dictionary iteration order. It now reports every bound pane, leader first, keeping the single-value `paneID` field as the leader so existing clients still read something sensible.
- **`--session` resolution was ambiguous for mirrors** and took `.first(where:)`. It now prefers the leader — the only defensible answer for a bare `$MACTERM_SESSION` self-target, since that variable lives in the session's single shared shell, which cannot know which of its views the user is looking at. Callers needing a specific mirror pass `--pane`.

## Verified

- [x] Built and ran the change in the app (`mise run e2e`, 24 passed)

Five unit tests plus one for the inference gate, and four new e2e against the real app and daemon: the source keeps leadership when mirrored; focus hands it over **and back** (it does not latch); `session list` reports both panes leader-first; `--session` follows the leader as focus moves.

## Notes for reviewers

**Known gap, documented in the code:** a foreign `zmx attach` from a real terminal can take leadership on its own — any keystroke does — without telling us, leaving the dim stale. Closing that needs a daemon-to-client notification zmx has no tag for, plus a ghostty change to surface it; not worth it for a case that only arises when someone attaches to a Macterm session from outside Macterm.

`paneIDsBySessionName` uses a **stable partition, not `sorted`** — "leader first" is not a strict weak ordering and Swift's sort is undefined for one.

### DIFF
--- a/CLI/MactermCommand.swift
+++ b/CLI/MactermCommand.swift
@@ -417,8 +417,9 @@ struct PaneCommand: ParsableCommand {
     /// boundary; this just hides the verb from `--help` in release.)
     private static var paneSubcommands: [ParsableCommand.Type] {
         var subs: [ParsableCommand.Type] = [
-            List.self, Inspect.self, Dump.self, Split.self, Focus.self,
-            Close.self, Run.self, Key.self, Zoom.self, ResizeSplit.self,
+            List.self, Inspect.self, Dump.self, Split.self, Mirror.self,
+            Focus.self, Close.self, Run.self, Key.self, Zoom.self,
+            ResizeSplit.self,
         ]
         #if DEBUG
         subs.append(Resize.self)
@@ -466,6 +467,24 @@ struct PaneCommand: ParsableCommand {
             args.direction = direction
             args.run = runCommand
             try runControlCommand(command: "pane.split", args: args, options: options)
+        }
+    }
+
+    struct Mirror: ParsableCommand {
+        static let configuration = CommandConfiguration(
+            abstract: "Mirror a pane — show the same session in a second pane."
+        )
+
+        @Option(help: "right, down, or auto (longer on-screen axis).")
+        var direction: String = "auto"
+
+        @OptionGroup var target: PaneTarget
+        @OptionGroup var options: ConnectionOptions
+
+        func run() throws {
+            var args = target.controlArgs()
+            args.direction = direction
+            try runControlCommand(command: "pane.mirror", args: args, options: options)
         }
     }
 
--- a/Macterm/App/AppState+PinnedTabs.swift
+++ b/Macterm/App/AppState+PinnedTabs.swift
@@ -298,8 +298,9 @@ extension AppState {
               let tab = ws.tabs.first(where: { $0.id == tabID })
         else { return }
         logger.info("unloadPinnedTab: \(tabID, privacy: .public)")
-        for pane in tab.splitRoot.allPanes() {
-            pane.killPersistentSession(using: zmx)
+        let unloading = tab.splitRoot.allPanes()
+        releaseSessions(unloading)
+        for pane in unloading {
             pane.destroySurface()
         }
         ws.closeTab(tabID)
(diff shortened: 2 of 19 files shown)


thdxg:
**Correction to the Verified section above.**

I ticked "built and ran the change in the app" on the grounds that `mise run e2e` drives the real GUI app. That was too loose: the e2e asserts JSON from `pane list`, so it proved the *leadership model*, not that the dim actually renders. The dim itself had not been looked at when I opened this.

It has now, and it works — but the first attempt measured **no dim at all**, and the reason is worth recording because it will mislead the next person who checks:

`unfocusedSplitFill` falls back to the theme background, so the overlay fades content *toward the background*. **Over an empty pane it is invisible by construction** — a pane showing only a prompt measured 25.07 vs 25.00 mean luminance, which reads as "the dim is broken" and isn't. You have to put content in the pane to see anything.

Re-measured with the pane full of text, capturing the app's own window (`screencapture -l<id>`) and comparing mean luminance of the two halves. Mirrors render byte-identical content, so any difference between the halves *is* the overlay:

```
leader = source   left=40.44  right=37.11   delta=+3.33
leader = mirror   left=36.07  right=41.96   delta=-5.89
```

The delta **inverts** when leadership moves, which is the part that rules out the two panes differing for some unrelated reason (a cursor, a prompt, scroll position).

Not added to the e2e suite: it needs Pillow and pyobjc-Quartz, which would be new CI dependencies for one visual assertion. The method above is enough to reproduce it by hand.

PR #356, opened 2026-09-07
Give each window its own adaptive terminal tint

thdxg:
> Stacked on #353. Review that first; the diff here is just this commit.

## What

The adaptive TUI background tint becomes **per window** instead of one app-wide value.

## Why

A visible bug that #353 introduces by existing. The tint was a single global, which was indistinguishable from per-window while Macterm had one window. With several, **every window wears whatever the focused one's terminal painted** — and changes colour again as focus moves, so a window showing a plain shell can sit there tinted by another window's editor.

Refs #345.

## How

The tint now belongs to the window it was sampled from — which is the rule `MactermTheme.nsConfiguredBg` **already encodes** for the quick terminal ("the panel must not paint the underlying window's tint"). That documented exception becomes the general case, and `WindowAppearance.sync` reads the window it is actually styling rather than a global.

**Sampling still follows only the key window**, as it already did — `monitoredViews()` has always picked one window. So a background window *keeps* the tint it was last sampled with rather than reverting to the theme: having nothing monitorable right now says nothing about what it is showing, and stripping the tint on focus-out would flicker every window switch.

Entries hold the window **weakly**, and are dropped when the delegate forgets the window, so a closed window cannot keep a tint alive for the life of the run.

`GhosttyApp.adaptiveBackgroundColor` survives as a **mirror** for the consumers that paint no particular window — a pane preview, a drag proxy, the tab switcher. It follows the key window's tint, falling back to the most recently sampled one, which is the window the sampler was following.

## Verified

- [x] Built and ran the change in the app (`mise run e2e`, 29 passed)
- [x] Added or updated tests

5 new unit tests. Sabotage-checked: reverting `effectiveBackgroundColor(for:)` to the app-wide value fails **three** of them, so they are not vacuous.

**The existing `ThemeTests` keep their teeth.** They pin that the scene-level `colorScheme` must *not* follow the tint — the thing that previously destabilized SwiftUI window management (Settings reopening on activation, hotkeys breaking on lost window identity). Under a per-window tint they would have passed trivially, satisfied by a nil global, so they now hang the tint on a window and still assert the real invariant.

## Notes for reviewers

**Not in this PR, and worth its own:** confirmation dialogs are still app-wide. `CloseConfirmationAlerts` is applied per window with ungated bindings, so a pending busy-close would raise the alert in **every** window — the same bug class the `DialogHost.settings` gate already documents for Settings.

The fix needs each alert modifier to know which window it is in, and they are applied to `MainWindow()` from the scene, *outside* where `WindowState` is injected. Moving them inside touches the alert grouping CLAUDE.md flags as a macOS 26.3 hazard ("add new alerts to one of these — never back into `body`"), so it deserves a deliberate change rather than being tacked on here. The shape I would use: `DialogHost.mainWindow` stays, staging records the key window's id, and each window renders only its own.

Sidebar width autosave is also still app-wide; that one is cosmetic (a restored width), and the plan notes the autosave name must stay **bounded** — per-window index, never a UUID — since unstable names are what caused the unbounded `NSSplitView Subview Frames` key accumulation the app already sweeps.


thdxg:
Added the two remaining per-window chrome items, both of which I'd flagged above as needing their own change.

## Sidebar width per window

It was one stored value *and* one static "already restored" flag — so only the **first** window ever had its width applied, and every other opened at SwiftUI's content-derived default. Each window now carries its own, persisted beside its project and restored per window.

**The autosave name is a reusable slot index, never a UUID.** One shared name would have windows fighting over the same AppKit key, but a per-window name has to stay *bounded*: an unstable one is exactly what produced the unbounded `NSSplitView Subview Frames` accumulation `pruneChurnedSidebarAutosaveKeys` exists to clean up. Slots are assigned lowest-free and reused when a window closes.

Dragging one window's divider still updates the app-wide default a **new** window opens at, so a new window matches what you just set.

**The clobber is removed rather than raced.** `launchSidebarWidth` existed because the column lays out — and the geometry hook writes its content-derived width over the stored value — before the window is styled. Instead of freezing a second per-window copy, the recorder now declines to write while a restore is still pending for that window. A nil `attachedWindow` counts as pending: the geometry hook fires during layout, *before* the styler has found the window, and treating nil as "nothing pending" is exactly how 144 kept landing in the snapshot — I watched it do so.

## Confirmation dialogs follow the focused window

Same shape as the Settings duplication the `DialogHost` gate already prevents: every window's scene carries its own copy of the alerts, so closing a busy pane with two windows open raised the warning in **both**. Each copy now presents only when its window is the one the user is in.

The modifiers moved from the scene into `MainWindow` to learn which window they belong to — they stay grouped in the same three modifiers, which is the rule (the alerts must not scatter back into `body`). `dialogWindowID` falls back to the first window when nothing is key, so a staged confirmation is never swallowed with nowhere to appear.

## Verified

Unit tests now 11 in `WindowStateTests` (independent widths, new-window default, dialog follows focus, dialog never has nowhere to go), plus one e2e.

**The restore path was verified by seeding a snapshot**, because neither a divider drag nor a second launch is reachable from the e2e suite. Two windows seeded at 210 and 330 came back as:

```
restored: [(1, 210), (2, 330)]
```

One honest limitation, now written into the e2e test: the harness isolates `$HOME` and the data dir but **not** UserDefaults (`Preferences.defaults` resolves through the user record), so the width a harness launch starts from is the developer's own debug-domain value. That is why the e2e asserts a range rather than a number — a fixed expectation there would be a machine-specific assertion. It also cost me a wrong conclusion first: I read a stored 144 as a fresh clobber before checking where it came from.

`mise run test` and `mise run e2e` (30 passed) are green locally, though the first e2e run right after a rebuild has flaked three times now on this machine — CI is the signal worth trusting.

PR #357, opened 2026-09-07
feat: give the sidebar more room by default

nabilfreeman:
## What

Makes the default sidebar 220 points wide instead of 180.

## Why

The current default is tight enough that ordinary project and tab names start fading early. A little more room keeps the sidebar readable without taking much away from the terminal.

Apple does not prescribe a fixed sidebar width. Its [sidebar](https://developer.apple.com/design/human-interface-guidelines/sidebars) and [split view](https://developer.apple.com/design/human-interface-guidelines/split-views) guidance focuses on reasonable pane sizes while keeping sidebars resizable and hideable. This keeps MacTerm's existing 140 to 400 point range, resizing, hiding, and width restoration unchanged.

## How

Only the fallback width changes. Fresh installs, or profiles without a saved width, start at 220 points.

An existing saved width remains authoritative, including 180 points if that is what someone chose.

## Verified

- [x] Added a regression assertion for the 220 point default
- [x] Launched an isolated profile with synthetic projects and confirmed the sidebar restored to 220 points
- [x] Compared the same labels at 180 and 220 points
- [x] Confirmed existing saved widths still pass through unchanged

## Notes for reviewers

Current 180 point default:

Current 180 point sidebar with several ordinary names fading

Proposed 220 point default:

Proposed 220 point sidebar with ordinary names shown in full

Full window at 220 points:

MacTerm with the proposed 220 point sidebar

### DIFF
--- a/Macterm/App/Preferences.swift
+++ b/Macterm/App/Preferences.swift
@@ -407,7 +407,7 @@ final class Preferences {
     /// `MainWindow`'s `navigationSplitViewColumnWidth` so a stored value can
     /// never fall outside what the column accepts.
     static let sidebarWidthRange: ClosedRange<Double> = 140 ... 400
-    static let defaultSidebarWidth: Double = 180
+    static let defaultSidebarWidth: Double = 220
 
     /// Which appcast channel auto-updates come from. Read by `Updater`'s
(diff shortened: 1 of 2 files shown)


thdxg:
@nabilfreeman LGTM!

PR #359, opened 2026-09-07
Harden multi-window: per-window tabs, close policy, leadership claims

thdxg:
Follow-up to the multi-window stack (#348–#356), from a review against the zmx and libghostty sources plus the tip.9 reports. Refs #345.

## What was wrong

**Reported on tip.9**
- Two windows on one project: the second drew blank. Both rendered `Workspace.activeTab`, and a pane's single `NSView` can live in one view hierarchy.
- ⌘N did nothing. A hardcoded "re-front the single window" fallback in `MainAppResponder` swallowed the chord before the File menu's New Window item saw it.
- The tab switcher popped up in every window and changed tabs everywhere.
- Closing a window, switching apps and coming back restored it. The red button hid every window (not only the last) and activation re-fronted the hidden one.
- Two windows did not come back after ⌘Q. Quit closes every window; each teardown unregistered and re-saved the snapshot with one window fewer.
- Sidebar collapse was shared across windows.

**Found in review**
- The leadership claim rides `ghostty_surface_key`, and libghostty treats any encoded key as typing: selection cleared, viewport scrolled to bottom (`Surface.zig` after encode). It was sent on every focus, so a double-click selection in a mirror vanished and a scrolled-up mirror jumped on each click. A zero keycode is also physical Key A on macOS.
- The 150ms trailing debounce let zmx drop the first click into a mouse-mode TUI and any arrow/Ctrl key typed right after a focus move: `handleInput` forwards a non-leader's bytes only when `isUserInput` says so, which excludes CSI mouse, bare Escape, plain arrows and Ctrl chords.
- Closing the leader left the survivor at the dead pane's pty size with its control keys dropped: zmx `closeClient` nils the leader and promotes nobody.
- Clicking a remote mirror recorded leadership the host never moved, so the dim lied.
- `window close` used `NSApp.keyWindow`, which is nil while the app is inactive, so a CLI call from another terminal closed the first window while `window list` said another was focused.
- Notification/CLI navigation repointed the key window even when another window already showed the project; unload/remove left other windows respawning the shells through their live surfaces.
- `nonLeaderPaneIDs` and `session list` were quadratic in pane count (a full traversal per pane, inside a view body).
- Restore used a single "next window" slot refilled by timers; per-window tables were keyed by `ObjectIdentifier`, whose address a later window can reuse.

## What changed

- **Per-window tab selection.** `WindowState.activeTabIDs` + `AppState.displayedTab(for:in:)` (key window first, then creation order, deterministic fallback). `Workspace.activeTabID` stays the key window's tab via `alignKeyWindowTab` and a workspace hook, so existing "the active tab" call sites are untouched. A window with nothing to show renders `TabShownElsewhereView`, whose action is `mirrorTab` (every pane mirrored, same structure) — this is how the same work gets onto two windows/Spaces. `window list` reports each window's `tabID`; `tab select`/`window close` take `--window`.
- **One close policy** (`AppDelegate.hidesInsteadOfClosing`): only the last visible window hides; red button, ⌘⇧W and CLI agree. `reopenIfNeeded` won't front a hidden window while one is visible. The window list is frozen during termination. Restore is a FIFO consumed at registration; the key window is recorded (`WindowSnapshot.isKey`) and fronted last.
- **Per-window presentation state**: sidebar visibility, palette, remote-project sheet, tab-switcher overlay. `AppState` keeps computed mirrors of the key window's copy for hotkeys/palette/CLI.
- **Claims**: sent only when leadership moves, synchronous, recorded only when delivered, unidentified keycode on text-only events, nothing sent or recorded for remote panes on focus (typed input is recorded for every pane from `keyDown` via `.terminalUserInput`). `releaseSessions` hands leadership to the surviving mirror when the leader closes.
- `revealProject` for navigation, `vacateWindows` on unload/remove, weak-keyed `NSMapTable`s and a single `windowDidClose` teardown, one-traversal session index, debounced window persistence.
- AGENTS.md "Windows (#345)" section replaces "Single Window Enforcement"; CLI docs cover the window verbs and `--window`.

## Verification

Unit suite green locally (17 new tests across `AppStateTests`/`WindowStateTests`, two existing tests adjusted to the new model). The e2e suite is left to CI: the new `test_two_windows_on_one_project_show_different_tabs` and two existing window tests passed in isolation but flaked in a full local run on a machine in active use, which is exactly the "first local e2e run flakes" the implementation report already records.

## Deliberately not done (follow-ups)

- Size mirrors to a common grid (tmux `window-size smallest`, doable client-side) so mirrors are never garbled and focus changes don't reflow.
- An out-of-band claim verb over the zmx daemon socket instead of the stdin APC (removes the filter from every user's keystroke path and the libghostty side effects).
- Declarative mirrors in layouts/pinned declarations — a saved layout still splits a mirror into two sessions.
- zmx reporting its leader client so a foreign `zmx attach` can't leave the dim stale; and a reserved high range for downstream zmx tag numbers.


thdxg:
Second round, from testing the branch:

- **Two windows on the same tab now mirror it** instead of one window getting a placeholder. The window that owns the tab keeps its real panes; every other window that selects it renders a shadow tab of mirror panes (same sessions, attached again), dimmed where the pty is sized for the leader. Ownership is sticky across key-window changes; the shadow rebuilds on shape changes and is dropped when the window moves on. Clicks, focus, zoom and splits in the mirror act on the real tab by tree position.
- **Tab selection no longer moves in the other window**: the fallback-shuffling that made a background window switch tabs when the key window took its tab is gone with the steal model.
- **⌘⇧W** was still handled by a single-window `orderOut(mainWindow)` in the responder, hiding the first window regardless of focus. That hidden window stayed registered and persisted, which is also why quitting with one window on screen brought two back. It now goes through `AppCommand.closeWindow`.
- The key-window callback could record a phantom (pre-adoption) `WindowState`; it now resolves through the `NSWindow`.
- `window list` reports `mirrored`; e2e test `test_two_windows_on_one_tab_mirror_it` covers the pair.

PR #360, opened 2026-09-07
Extend the tab switcher with adaptive preview cards

Pavior0:
## What

Follow-up to #346.

Extend the existing Recent Tab switcher with a separate panel, preview-sized cards, pointer selection, and a configurable candidate count.

The candidate count defaults to 5 recent tabs. Users can choose 2 to 12 tabs or Unlimited.

## Why

The current switcher lays out every recent tab in one horizontal strip inside the terminal window.

That works for a small number of tabs, but the cards become narrow as the list grows. The preview and its hit area are both difficult to use with the pointer. Once the strip overflows, the user has to keep pressing Recent Tab to reach cards outside the visible area. There is no useful mouse or trackpad path through the overflow.

The fixed card shape also does not match every terminal layout. Split panes and narrow panes can produce cards with too little room for their contents.

## How

The switcher now renders in a borderless, non-activating `NSPanel`. The panel can extend beyond the terminal window, but stays inside the visible screen. It does not take keyboard focus, so releasing the Recent Tab modifier still commits through the existing key handling.

Cards use the aspect ratio of their captured terminal composition. Split panes are composed recursively, previews share a 104pt height, and card widths are bounded between 150pt and 220pt. The preview uses `.fit`, so tall and split layouts are not cropped. Cards without a captured frame keep the existing text fallback.

The default candidate count keeps the common case compact. Users can choose 2 to 12 recent tabs or Unlimited in Settings. Unlimited mode lays cards out in rows and scrolls vertically, so the mouse and trackpad can reach cards outside the first row.

The whole card is interactive. Hovering moves the pending selection, clicking commits it, and releasing the modifier commits the highlighted card. Escape and outside clicks cancel the cycle. When the overlay is disabled, direct cycling keeps the existing full recency order and behavior.

## Verified

- [x] Built and ran the change in the app and confirmed the behavior

## Notes for reviewers

The state and preference changes are covered in `AppStateTests` and `PreferencesTests`.

| Before | After |
| --- | --- |
| Before 1<br><br>Before 2 | After 1<br><br>After 2 |


thdxg:
@Pavior0 Thanks for this. I like the idea of wrapping overflow preview tabs and limiting the number of candidates. But i want to clarify some things:
- I think previews should respect the current aspect ratio of the tab. Your version seems to have a fixed ratio for previews.
- Pointer selection already works. Not sure which part of the pointer selection this PR addresses.

Also, I am planning to implement live previews in the tab switcher, allowing the previews to update real-time as terminal outputs change. I'd like to defer UI changes to after implementing that feature.

Edit: I the live preview feature has been merged. Would you be willing to set the scope of this PR to wrapping overflowing tab previews and limiting candidates?

Pavior0:
Closing this in favor of #376. The new PR follows up with a much smaller scope focused on wrapping overflowing previews and limiting the candidate count, based on the current `main`.

PR #376, opened 2026-09-08
Wrap overflowing previews in the Tab Switcher and limit candidates

Pavior0:
## What

Follow-up to #360.

The Recent Tab switcher now wraps previews into centered rows. A new setting limits the candidate count to 2 through 12 tabs, or Unlimited. The default is 5.

When the rows exceed the window height, the card area scrolls vertically and keeps the selected row in view.

## Why

The current switcher shows a moving horizontal slice when its previews no longer fit. Tabs outside that slice stay hidden until the selection reaches them.

Wrapping places the candidates into rows. The default limit keeps the switcher compact, while vertical scrolling makes Unlimited usable without letting the panel grow past the window.

## How

The switcher calculates how many preview cards fit at the current window width, then divides the candidates into centered rows.

The panel keeps its natural height when all rows fit. Otherwise, it uses a bounded vertical scroll view with a system scroll bar.

Keyboard cycling centers the selected row with a short animation. Hover selection does not move the scroll position, and Reduce Motion disables the animation.

The candidate limit applies only when the overlay is enabled. Unlimited and direct cycling without the overlay continue to use the full recency order.

## Verified

- [x] Built and ran the change with `mise run run`
- [x] Confirmed that five previews wrap as 2 + 2 + 1 in a narrow window
- [x] Confirmed that overflow stays inside the window and exposes a vertical scroll bar
- [x] Confirmed that cycling keeps the selected row centered with its full previews visible
- [x] Confirmed that terminal previews continue updating while the switcher is open
- [x] Added tests for candidate limiting and preference persistence


thdxg:
@Pavior0 Thanks for your work! I made some minor fixes. This will be on the next release!

PR #397, opened 2026-09-16
Fix Split Left and Split Up, and give the CLI the same sides

Pavior0:
## What

Split Left and Split Up now put the new pane left and above. `pane split` and
`pane mirror` take `--direction left` and `up` too.

## Why

`SplitTreeView` dropped `SplitPosition` and `TerminalTab.split` hardcoded
`.second`, so the menu offered four sides and two of them were dead. Drag and
drop read the value correctly, which is why nobody reported it. No upstream
issue covers this.

## How

Thread `SplitPosition` through the views, `AppState`, and `TerminalTab` down
to `SplitNode.splitting`, which already supported both sides. `TerminalTab.mirror`
takes it too, since the CLI parses `--direction` once for both verbs.

No hotkey or `AppCommand` for left and up. That needs a `MactermKeybind` case,
so it is its own change.

## Verified

## Notes for reviewers

- I checked left, up, right, and a bad direction through the CLI on the debug
  build.
- Commit 1 is the menu fix, commit 2 the CLI.
- The palette and keybinds still list only Right, Down, and Automatically. Left
  and up are a follow-up so this PR stays a fix


thdxg:
@Pavior0 Thanks! LGTM

